PR #3, opened 2022-06-29
Global attr setup

hashedone:


hashedone:
Abandoning this approach: https://github.com/CosmWasm/sylvia/issues/6#issuecomment-1215424024

PR #11, opened 2022-08-16
Adding coverage for cw1 sylvia messages

jawoznia:


contracts/cw1/Cargo.toml
```diff
@@ -0,0 +1,33 @@
...
+edition = "2021"
+description = "todo"
+license = "Apache-2.0"
+repository = "https://github.com/CosmWasm/sylvia"
+homepage = "https://cosmwasm.com"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[lib]
+crate-type = ["cdylib", "rlib"]
```
hashedone, line 14: This one is not a complete contract, it doesn't need to be `cdylib`. `cdylib` is only for WASM binaries, so final contracts, anything else is just rlib.

contracts/cw1/Cargo.toml
```diff
@@ -0,0 +1,33 @@
...
+cw-utils = "0.14.0" 
+cw1 = "0.14.0" 
+cw2 = "0.14.0" 
+cosmwasm-std = { version = "1.0.0", features = ["staking"] }
+cw-storage-plus = "0.14.0" 
+schemars = "0.8.1"
+serde = { version = "1.0.103", default-features = false, features = ["derive"] }
+thiserror = { version = "1.0.23" }
+sylvia = { version = "0.1.2", path = "../../sylvia" }
+sylvia-derive = {version = "0.1", path = "../../sylvia-derive" }
```
hashedone, line 26: I don't want any contracts to depend on `sylvia-derive` which is meant to be an only implementation detail. Everything is (or should be - if not, correct it) exposed via `sylvia` crate. If you have to include `sylvia-derive` in contracts/contract interfaces, it means it should be fixed in the framework.

contracts/cw1/Cargo.toml
```diff
@@ -0,0 +1,33 @@
...
+crate-type = ["cdylib", "rlib"]
+
+[dependencies]
+cw-utils = "0.14.0" 
+cw1 = "0.14.0" 
+cw2 = "0.14.0" 
+cosmwasm-std = { version = "1.0.0", features = ["staking"] }
+cw-storage-plus = "0.14.0" 
+schemars = "0.8.1"
+serde = { version = "1.0.103", default-features = false, features = ["derive"] }
```
hashedone, line 23: Also, I think you should get rid of serde - it is reexposed by Sylvia, and it would be good to use `sylvia::serde` in place of `serde` to have version compatibility guaranteed. Same for `schemars`.

contracts/cw1/Cargo.toml
```diff
@@ -0,0 +1,33 @@
...
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[lib]
+crate-type = ["cdylib", "rlib"]
+
+[dependencies]
+cw-utils = "0.14.0" 
+cw1 = "0.14.0" 
+cw2 = "0.14.0"
```
hashedone, line 19: This is meant to be purely interfaces, why do you need those?

contracts/cw1/src/lib.rs
```diff
@@ -0,0 +1,50 @@
...
+        ctx: (DepsMut, Env, MessageInfo),
+        member: String,
+    ) -> Result<Response, Self::Error>;
+
+    #[msg(query)]
+    fn find_member(&self, ctx: (Deps, Env), member: String) -> Result<Response, Self::Error>;
+}
+
+#[cfg(test)]
+mod tests {
```
hashedone, line 22: I see you are testing if serialization->deserialization is an identity function which is good, but it is a bit missing here. You want to make sure, that the in-between product is proper. In particular, you want to make sure, that deserialization of JSONs you expect is proper.

Testing the serialized JSONs would be a bit strange, as you will need to consider formatting, and so, and maintaining it would be hell. What I would do is deserialize hand-crafted JSON messages (with `cosmwasm_std::from_slice`).

contracts/cw1/Cargo.toml
```diff
@@ -0,0 +1,33 @@
...
+thiserror = { version = "1.0.23" }
+sylvia = { version = "0.1.2", path = "../../sylvia" }
+sylvia-derive = {version = "0.1", path = "../../sylvia-derive" }
+
+[dev-dependencies]
+anyhow = "1"
+assert_matches = "1"
+cosmwasm-schema = { version = "1.0.0" }
+cw-multi-test = "0.14.0"
+derivative = "2"
```
hashedone, line 33: Did you use any of those?

contracts/cw1/Cargo.toml
```diff
@@ -0,0 +1,33 @@
...
+
+[lib]
+crate-type = ["cdylib", "rlib"]
+
+[dependencies]
+cw-utils = "0.14.0" 
+cw1 = "0.14.0" 
+cw2 = "0.14.0" 
+cosmwasm-std = { version = "1.0.0", features = ["staking"] }
+cw-storage-plus = "0.14.0"
```
hashedone, line 21: Also no state in this thing I think

contracts/cw1/Cargo.toml
```diff
@@ -0,0 +1,33 @@
...
+
+[dependencies]
+cw-utils = "0.14.0" 
+cw1 = "0.14.0" 
+cw2 = "0.14.0" 
+cosmwasm-std = { version = "1.0.0", features = ["staking"] }
+cw-storage-plus = "0.14.0" 
+schemars = "0.8.1"
+serde = { version = "1.0.103", default-features = false, features = ["derive"] }
+thiserror = { version = "1.0.23" }
```
hashedone, line 24: Also redundant

hashedone (commented):
In general Ok, some cleanup is needed. Plus a bit of testing if messages as what we actually expect them to be (not only symmetrically serializable).

contracts/cw1/Cargo.toml
jawoznia: As discussed offline this has to stay as sylvia-derive from which interface macro is being used requires it.

contracts/cw1/Cargo.toml
```diff
@@ -0,0 +1,15 @@
...
+authors = ["Jan Woźniak <jan@confio.gmbh>"]
+edition = "2021"
+description = "Coverage of cw1 sylvia message generation."
+license = "Apache-2.0"
+repository = "https://github.com/CosmWasm/sylvia"
+homepage = "https://cosmwasm.com"
+
+[dependencies]
+cosmwasm-std = { version = "1.0.0", features = ["staking"] }
+schemars = "0.8.1"
```
hashedone, line 13: I assume you checked it is required? It might be for the same reason as serde is, but if it would work without it then remove (I am not sure tbh).

hashedone (approved):
lgtm

contracts/cw1/Cargo.toml
jawoznia: Unfortunately it failed with the same reason as serde.

PR #12, opened 2022-08-19
Adding Eq trait in few places with clippy.

jawoznia:
There are still some warnings with lack of Eq in interface
generated msgs. Unfortunately adding Eq there is not allowed
due std::cmp::Eq not being implemented for cosmwasm::CosmosMsg.


hashedone (approved):
lgtm

PR #13, opened 2022-08-19
Cw1 whitelist

jawoznia:
Coverage of Cw1Whitelist in sylvia.

### DIFF
--- a/contracts/cw1-whitelist/Cargo.toml
+++ b/contracts/cw1-whitelist/Cargo.toml
@@ -16,3 +16,4 @@ sylvia = { version = "0.1.2", path = "../../sylvia" }
 cw1 = { version = "0.1.2", path = "../cw1" }
 cw-storage-plus = "0.13.2"
 anyhow = "1"
+cw-multi-test = "0.14.0"
--- a/contracts/cw1-whitelist/Cargo.toml
+++ b/contracts/cw1-whitelist/Cargo.toml
@@ -16,5 +16,8 @@ sylvia = { version = "0.1.2", path = "../../sylvia" }
 cw1 = { version = "0.1.2", path = "../cw1" }
 cw-storage-plus = "0.13.2"
 thiserror = { version = "1.0.23" }
+
+
+[dev-dependencies]
 anyhow = "1"
 cw-multi-test = "0.14.0"
--- a/contracts/cw1-whitelist/Cargo.toml
+++ b/contracts/cw1-whitelist/Cargo.toml
@@ -15,5 +15,6 @@ serde = { version = "1", default-features = false, features = ["derive"] }
 sylvia = { version = "0.1.2", path = "../../sylvia" }
 cw1 = { version = "0.1.2", path = "../cw1" }
 cw-storage-plus = "0.13.2"
+thiserror = { version = "1.0.23" }
 anyhow = "1"
 cw-multi-test = "0.14.0"
--- a/contracts/cw1-whitelist/Cargo.toml
+++ b/contracts/cw1-whitelist/Cargo.toml
@@ -1,0 +1,18 @@
+[package]
+name = "cw1-whitelist"
+version = "0.1.2"
+authors = ["Jan Woźniak <jan@confio.gmbh>"]
+edition = "2021"
+description = "Coverage of cw1-whitelist sylvia message generation."
+license = "Apache-2.0"
+repository = "https://github.com/CosmWasm/sylvia"
+homepage = "https://cosmwasm.com"
+
+[dependencies]
+cosmwasm-std = { version = "1.0.0", features = ["staking"] }
+schemars = "0.8.1"
+serde = { version = "1", default-features = false, features = ["derive"] }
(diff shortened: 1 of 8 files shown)


contracts/cw1-whitelist/Cargo.toml
```diff
@@ -0,0 +1,19 @@
...
+
+[dependencies]
+cosmwasm-std = { version = "1.0.0", features = ["staking"] }
+schemars = "0.8.1"
+serde = { version = "1", default-features = false, features = ["derive"] }
+sylvia = { version = "0.1.2", path = "../../sylvia" }
+cw1 = { version = "0.1.2", path = "../cw1" }
+cw-storage-plus = "0.13.2"
+anyhow = "1"
+cw-multi-test = "0.14.0"
```
hashedone, line 19: Those should be used in tests only, which means you want those in `[dev-dependencies]`

contracts/cw1-whitelist/src/contract.rs
```diff
@@ -0,0 +1,198 @@
...
+use cw1::msg::{ExecMsg, QueryMsg};
+use cw1::Cw1;
+use cw1::*;
+use cw_storage_plus::Map;
+use sylvia::contract;
+
+use self::contract::InstantiateMsg;
+
+pub struct Cw1WhitelistContract {
+    members: Map<'static, Addr, Empty>,
```
hashedone, line 12: I think we want this to be public - if another contract would like to extend this one it would probably want to have access to its state. However it decreases hermetization, and it's not clear how to approach this, we will build some proper patterns in future so we can leave it like it is, for now, will see what works best when we will add more contracts.

contracts/cw1-whitelist/src/contract.rs
```diff
@@ -0,0 +1,198 @@
+use anyhow::Error;
```
hashedone, line 1: Don't use `anyhow` on production, its a bit more overhead than `thiserror` - better switch to the second one. It also gives you better error testability.

contracts/cw1-whitelist/src/contract.rs
```diff
@@ -0,0 +1,198 @@
...
+        msg: &[u8],
+    ) -> Result<Response, Error> {
+        let msg: ExecMsg = from_slice(msg)?;
+        msg.dispatch(self, (deps, env, info))
+    }
+
+    pub(crate) fn entry_query(&self, deps: Deps, env: Env, msg: &[u8]) -> Result<Binary, Error> {
+        let msg: QueryMsg = from_slice(msg)?;
+        msg.dispatch(self, (deps, env))
+    }
```
hashedone, line 93: I know I did this in the original `cw1-whitelist`, but I feel this is a very nasty boilerplate. Those functions adds more complexity than just calling `from_slice::<ExecMsg>(msg)?.dispatch(CONTRACT, (deps, env, info))` in the entry point - I think we should prefer this approach. I know you would double it in `Contract` implementation, but this one is to be generated anyway so you will not see that.

contracts/cw1-whitelist/src/lib.rs
```diff
@@ -0,0 +1,52 @@
...
+
+    #[entry_point]
+    pub fn query(deps: Deps, env: Env, msg: Binary) -> Result<Binary, Error> {
+        CONTRACT.entry_query(deps, env, &msg)
+    }
+
+    #[entry_point]
+    pub fn sudo(_deps: DepsMut, _env: Env, _msg: Vec<u8>) -> AnyResult<Response> {
+        bail!("sudo not implemented for contract")
+    }
```
hashedone, line 41: You don't need nor want that entry point. Bail directly in `Contract` implementation. Here they would end up in the binary for no reason. (same for `reply` and `migrate`)

contracts/cw1/src/lib.rs
```diff
@@ -15,7 +22,11 @@ pub trait Cw1 {
     ) -> Result<Response, Self::Error>;
 
     #[msg(query)]
-    fn find_member(&self, ctx: (Deps, Env), member: String) -> Result<Response, Self::Error>;
```
hashedone, line 18: Ahh, that was factually wrong. Good you corrected it.

hashedone (approved):
LGTM, with some comments - please correct them before merging. I see many places to improve on sylvia to make it better :D

contracts/cw1-whitelist/src/contract.rs
```diff
@@ -0,0 +1,171 @@
+use crate::error::ContractError;
+use cosmwasm_std::{Addr, DepsMut, Empty, Env, MessageInfo, Response};
+
+use cw1::Cw1;
+use cw1::*;
+use cw_storage_plus::Map;
+use sylvia::contract;
+
+pub struct Cw1WhitelistContract {
+    members: Map<'static, Addr, Empty>,
```
ethanfrey, line 10: Actually, by biggest nit is use by value not reference. Please use:

`members: Map<'static, &Addr, Empty>,`

(And I think you can skip `'static` maybe? if it compiles without)

PR #23, opened 2022-08-23
Print messages

### LINKED ISSUE #7: Generate the `const fn messages() -> [&'static str; N]` function returning generated messages variants
The function should be generated for both query and exec messages. The idea is to be able to use them to verify if message variants collide, and do it in compile time. For example, consider the following trait:

```rust
#[sylvia::interface]
pub trait Group {
  type Error: From<StdError>;

  #[msg(exec)]
  fn update_admin(&self, ctx: ExecCtx, admin: Option<String>) -> Result<Response, Self::Error>;
  
  #[msg(exec)]
  fn update_members(&self, ctx: ExecCtx, remove: Vec<String>, add: Vec<Member>) -> Result<Response, Self::Error>;
  
  #[msg(query)]
  fn member(&self, ctx: QueryCtx, addr: String) -> Result<MemberResp, Self::Error>;
}
```

The following should be generated:

```rust
impl ExecMsg {
  const fn messages() -> [&'static str; 2] {
    ["update_admin", "update_members"]
  }
}

impl QueryMsg {
  const fn messages() -> [&'static str; 1] {
    ["member"]
  }
}
```

Note, that it is probably beneficial to keep all items sorted internally (so it would be linear to look for name collisions).

jawoznia:
Print messages generated by Sylvia.
close #7


contracts/cw1-whitelist/src/contract.rs
```diff
@@ -27,6 +27,21 @@ impl Cw1 for Cw1WhitelistContract {
         Ok(Response::new())
     }
 
+    fn remove_member(
```
hashedone, line 30: `cw1` does not contain `remove_member`. If you want to try msg with multiple alternatives, just create `cw4`

hashedone (commented):
Just make the test better aligned to reality (leave cw1 being cw1, and for multiple messages, variants add cw4 for example).

contracts/cw1/src/lib.rs
```diff
@@ -97,6 +90,6 @@ mod tests {
 
     #[test]
     fn exec_msgs() {
-        assert_eq!(ExecMsg::messages(), ["AddMember", "RemoveMember"]);
+        assert_eq!(ExecMsg::messages(), ["AddMember"]);
```
hashedone, line 93: I didn't notice it before - you want it to be "add_member", not "AddMember" - it would be way useful, as it would be always used with incoming jsons. Check out the `convert_case` crate (already in dependencies).

contracts/cw4/Cargo.toml
```diff
@@ -0,0 +1,15 @@
...
+description = "Coverage of cw4 sylvia message generation."
+license = "Apache-2.0"
+repository = "https://github.com/CosmWasm/sylvia"
+homepage = "https://cosmwasm.com"
+
+[dependencies]
+cosmwasm-std = { version = "1.0.0", features = ["staking"] }
+schemars = "0.8.1"
+serde = { version = "1", default-features = false, features = ["derive"] }
+sylvia = { version = "0.1.2", path = "../../sylvia" }
```
hashedone, line 15: I don't like this missing newline

hashedone (commented):
Another change request - sorry I missed it in first iteration

Issue #10, opened 2022-08-15
Add a possibility to define `#[exec]` and `#[query]` messages directly on contract type


hashedone:
Right now, all query and execute messages have to be implemented by traits. It is good for interface reusing, but sometimes we want to have very contract-specific messages we do not expect to be implemented on any other contracts. For such cases, it would be a nice option to define `#[exec]` and `#[query]` directly on the contract.

Implementation wise - it would generate additional message structure, as they would be defined on some additional traits.

Issue #16, opened 2022-08-22
Add support for custom queries


hashedone:
At the moment there is a possibility to set the custom message with `msg_type`. `query_type` should be added to overwrite the `CustomQuery` so the dispatch is generated properly.

PR #24, opened 2022-08-25
Allow defining #exec and #query msgs in contract impl.

jawoznia:


hashedone:
Close #10 

Please remember about putting that in the description - it helps relate PRs to issues.

sylvia-derive/src/message.rs
```diff
@@ -609,11 +803,14 @@ impl<'a> GlueMessage<'a> {
...
+                    #impl_deserialization_attempt
 
                     Err(D::Error::custom(format!(
-                        "Expected any of {} messages, but cannot deserialize to neither of those\n{}",
+                        "Expected any of {} or {} messages, but cannot deserialize to neither of those\n{}{}",
                         stringify!(#(#messages_type),*),
-                        [#(#deser_errors,)*].join("\n")
+                        stringify!(#impl_msg_name),
+                        [#(#deser_errors,)*].join("\n"),
+                        [#impl_deser_errors].join("\n")
```
hashedone, line 813: I am not a fan of the fact, that you are adding a new argument there. You can just add `impl_deser_errors` to the first array:

```suggestion
                        stringify!(#impl_msg_name),
                        [#(#deser_errors,)* #impl_deser_errors].join("\n")
```

This may be a bit cheaper, as you are joining on the single array (in particular - the second join is redundant as it is always one element array!). Obviously you also need to remove one `{}` from format string.

sylvia-derive/src/parser.rs
```diff
@@ -63,6 +67,7 @@ impl Parse for ContractArgs {
     fn parse(input: ParseStream) -> Result<Self> {
         let mut module = None;
         let mut error = parse_quote!(ContractError);
+        let mut msg_type = None;
```
hashedone, line 70: Remove it right now, we will make it properly in #16 together with queries.

hashedone (commented):
Additionally - the test doesn't pass `cargo fmt --check`. Format all your code with `cargo fmt`. And add `cargo fmt --check` to the pipeline please.

contracts/cw1-whitelist/src/contract.rs
```diff
@@ -100,6 +119,54 @@ mod tests {
         );
     }
 
+    #[test]
+    fn binary_serialize_exec() {
+        let original_msg = ExecMsg::Cw1WhitelistContract(ImplExecMsg::RemoveMember {
```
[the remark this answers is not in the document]
jawoznia: I agree this is bad. This is definitely for future refactor consideration.

sylvia-derive/src/parser.rs
```diff
@@ -22,6 +22,10 @@ pub struct ContractArgs {
     pub module: Option<Ident>,
     /// The type of a contract error for entry points - `ContractError` by default
     pub error: Type,
+    /// The type being a parameter of `CosmosMsg` for blockchain it is intendet to be used; can be
+    /// set to any of generic parameters to create interface being generic over blockchains; If not
+    /// provided, cosmos messages would be unparametrized (so default would be used)
+    pub msg_type: Option<Type>,
```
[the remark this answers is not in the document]
jawoznia: Removed as proposed in further comment. It was added on development stage and I didn't notice it is not needed anymore.

contracts/cw1-whitelist/src/contract.rs
```diff
@@ -62,6 +68,29 @@ impl Cw1WhitelistContract {
...
+        self.members
+            .remove(deps.storage, deps.api.addr_validate(&member)?);
+
+        Ok(Response::new()
+            .add_attribute("action", "remove_member")
+            .add_attribute("sender", msg.sender))
+    }
+
+    #[msg(query)]
+    fn members_cnt(&self, (deps, _env): (Deps, Env)) -> Result<MembersCntResponse, ContractError> {
```
hashedone, line 87: That makes way more sense

contracts/cw1-whitelist/src/multitest.rs
```diff
@@ -92,4 +92,45 @@ mod test {
 
         assert_eq!(resp, FindMemberResponse { is_present: true });
     }
+
+    #[test]
+    fn contract_exec() {
+        let mut app = AppBuilder::new().build(|_, _, _| ());
```
hashedone, line 98: Donald use `AppBuilder` if you are using default `App`. Just use `App::default()`.

sylvia-derive/src/message.rs
```diff
@@ -552,11 +725,19 @@ impl<'a> GlueMessage<'a> {
             quote! { #variant(#module :: #name<#(#generics,)*>) }
         });
 
+        let impl_msg_name = match msg_ty {
+            MsgType::Exec => Ident::new("ImplExecMsg", Span::call_site()),
+            MsgType::Query => Ident::new("ImplQueryMsg", Span::call_site()),
+            MsgType::Instantiate => todo!(),
```
hashedone, line 731: Why panicking with todo? Handle the case - possibly by emitting an error.

sylvia-derive/src/message.rs
```diff
@@ -609,11 +803,13 @@ impl<'a> GlueMessage<'a> {
...
                     #(#deserialization_attempts)*
+                    #impl_deserialization_attempt
 
                     Err(D::Error::custom(format!(
-                        "Expected any of {} messages, but cannot deserialize to neither of those\n{}",
+                        "Expected any of {} or {} messages, but cannot deserialize to neither of those\n{}",
                         stringify!(#(#messages_type),*),
-                        [#(#deser_errors,)*].join("\n")
+                        stringify!(#impl_msg_name),
+                        [#(#deser_errors,)* #impl_deser_errors].join("\n")
```
hashedone, line 812: Why didnt you just add a message to a variant array as I suggested? Seems like IT world reduce generated code a bit.

hashedone (commented):
In general ok, left some comments. When I am back home I will double chęci the generated output as I do tego have toolset on my Phone.

sylvia-derive/src/message.rs
jawoznia: https://github.com/CosmWasm/sylvia/pull/24#discussion_r957698212

I'm pretty sure that it's the same as in your suggestion. Please take a look once more.

hashedone: Yea, it is the same. I was misreading on phone. Sorry.

contracts/cw1-whitelist/src/contract.rs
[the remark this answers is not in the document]
hashedone: Yup, we agreed on this solution - definitely do not try to solve it now, it requires deeper think how to approach this. Just mentioning.

sylvia-derive/src/message.rs
```diff
@@ -609,11 +809,13 @@ impl<'a> GlueMessage<'a> {
...
 
                     #(#deserialization_attempts)*
+                    #impl_deserialization_attempt
 
                     Err(D::Error::custom(format!(
-                        "Expected any of {} messages, but cannot deserialize to neither of those\n{}",
+                        "Expected any of {} or {} messages, but cannot deserialize to neither of those\n{}",
                         stringify!(#(#messages_type),*),
-                        [#(#deser_errors,)*].join("\n")
+                        stringify!(#impl_msg_name),
```
hashedone, line 817: Here you could add this `#impl_msg_name` to the previously `stringify`, but its not a problem I think

hashedone (approved):
LGTM at this point

contracts/cw1-whitelist/src/contract.rs
```diff
@@ -62,6 +68,29 @@ impl Cw1WhitelistContract {
 
         Ok(Response::new())
     }
+
+    #[msg(exec)]
+    fn remove_member(
+        &self,
+        (deps, _env, msg): (DepsMut, Env, MessageInfo),
```
ethanfrey, line 75: `(deps, _env, msg): (DepsMut, Env, MessageInfo),`

This reads a bit ugly / log. Especially elsewhere where we have `cosmwasm_std::DepsMut`.
If we define some struct like `type CtxMut = (DepsMut, Env, MessageInfo)` that could make this read cleaner (and maybe fit on one line).

I do like the destructuring in the arg list

contracts/cw1-whitelist/src/contract.rs
```diff
@@ -62,6 +68,29 @@ impl Cw1WhitelistContract {
...
+        member: String,
+    ) -> Result<Response, ContractError> {
+        self.members
+            .remove(deps.storage, deps.api.addr_validate(&member)?);
```
ethanfrey, line 79: Using &Addr as Map key could be more efficient.
For example, if we used msg.sender here (which is also used below in the events emitted), we would need:

`self.members.remove(deps.storage, msg.sender.clone());`

With reference, only:

`self.members.remove(deps.storage, &msg.sender);`

And they both accomplish the same work under the hood. One just forces unnecessary clone

contracts/cw1-whitelist/src/contract.rs
```diff
@@ -100,6 +129,49 @@ mod tests {
...
+
+        let serialized_msg = to_binary(&original_msg).unwrap();
+        let serialized_msg: QueryMsg = from_binary(&serialized_msg).unwrap();
+
+        assert_eq!(serialized_msg, original_msg);
+    }
+
+    #[test]
+    fn slice_deserialize_query() {
+        let deserialized: QueryMsg = from_slice(br#"{"members_cnt": {}}"#).unwrap();
```
ethanfrey, line 168: This is great json serialization (except for use of "cnt" rather than "count")

contracts/cw1-whitelist/src/contract.rs
```diff
@@ -100,6 +129,49 @@ mod tests {
...
+            deserialized,
+            ExecMsg::Cw1WhitelistContract(ImplExecMsg::RemoveMember {
+                member: "some_member".to_owned()
+            })
+        );
+    }
+
+    #[test]
+    fn binary_serialize_query() {
+        let original_msg = QueryMsg::Cw1WhitelistContract(ImplQueryMsg::MembersCnt {});
```
ethanfrey, line 158: Wondering if the JsonSchema is properly generated for this type?
Does it show up flattened or nested?

I think there is no schema.rs file yet (or is that auto-generated)? But something to keep in mind, that it must also flatten these nested structs.

ethanfrey: Also, using this is code could be annoying if you must write for many test cases. I understand the limits of codegen, but have two possible ideas, maybe one is feasible:

* `QueryMsg::MembersCnt{}` (actually flatten in the proc macros - likely hard)
* `ImplQueryMsg::MembersCnt {}.into()` (just add a helper to wrap it into QueryMsg)

Both are much less typing. Open for other ideas

hashedone: Actually good point, we didn't check it. However I expect it does, as we are adding `#[serde(untagged)]` for proper serialization, and `schemars` should read all serde tags. However - will double-check.

For executing query in real code, you don't use `QueryMsg::Cw1WhitelistContract(ImplQueryMsg::MembersCnt {})` anyway, it is for testing purposes to ensure uniform serialization. Normally you would use `ImplQueryMsg::MembersCnt {}` as its serialization is guaranteed to be the same (and you are working on JSON API-wise). We would prefer to go with `QueryMsg::MembersCnt{}` instead. but here is a problem - it doesn't go well with `#[serde(untagged)]`. We are figuring out the best namings scheme - one of ideas is to rename the final contract messages (as they should almost never be used) to some `GlueQueryMsg` and the contract-specific to just `QueryMsg` so it is very natural call (in this very test it would be `GlueQueryMsg::Cw1WhitelistContract(QueryMsg::MembersCnt {})`, but in real-live query just `QueryMsg::MembersCnt {}`).

PR #33, opened 2022-08-30
Allow derive partial eq without eq and CI coverage

### LINKED ISSUE #31: Align generated code to the new clippy lint
Generated code fails on `clippy::derive_partial_eq_without_eq` - the easiest way here should be allow the warning. Generated messages could have `Eq` generated, but it does not work for cases when using some args with no `Eq` on them (eg. Addr).

jawoznia:
close #31

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,48 @@
+name: Contract Development
+
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+    branches:
+      - main
+
+env:
+  RUST_BACKTRACE: 1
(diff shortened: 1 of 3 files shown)


hashedone:
No, we  dont want to allow warnings in contracts. It means Sylwia user would have to do that. We want to generate those allows for the generated stricte directly. And we want to remove allowance of deadcode.

Albo add in CI calling cw-check-contract on smart contracts.

jawoznia:
> 

This is being fixed under #24

hashedone (approved):
LGTM

Issue #30, opened 2022-08-26
Introduce custom context types


hashedone:
All functions in sylvia take some context on the first value - for now, those are typically tuples of cosmwasm-std types, but it makes using them weird. I propose to define context type for all kinds of messages, for example:
```rust
struct ExecContext {
  deps: DepsMut,
  env: Env,
  info: MessageInfo,
}
```
So it is a bit less noise. It is very easy to implement, as sylvia always calls `.into()` on passed tuples, so all those context structures need is to implement `From<...>` trait.

PR #39, opened 2022-09-08
Cw1 Whitelist contracts done properly

hashedone:


contracts/cw1-whitelist/src/contract.rs
```diff
@@ -1,243 +1,451 @@
...
+            mutable: Item::new("mutable"),
         }
     }
 
     #[msg(instantiate)]
     pub fn instantiate(
         &self,
-        (deps, _env, _msg): (DepsMut, Env, MessageInfo),
-        members: Vec<String>,
+        ctx: (DepsMut, Env, MessageInfo),
```
jawoznia, line 30: Maybe just `(deps, _, _): (DepsMut, Env, MessageInfo),`

I know it's not the preatiest approach, but you do it later in line 34 anyway.
Until we will introduce #30 I think we can keep it as it was with only the difference to change `_env` to just `_`

contracts/cw1-whitelist/src/contract.rs
```diff
@@ -1,243 +1,451 @@
...
+        mutable: bool,
     ) -> Result<Response, ContractError> {
-        for addr in members.into_iter() {
-            self.members
-                .save(deps.storage, deps.api.addr_validate(&addr)?, &Empty {})?;
+        let (deps, _, _) = ctx;
+        set_contract_version(deps.storage, CONTRACT_NAME, CONTRACT_VERSION)?;
+
+        for admin in admins {
+            let admin = deps.api.addr_validate(&admin)?;
```
jawoznia, line 38: Just a nitpick to consider. Bot versions fine by me
```
            self.admins
                .save(deps.storage, &deps.api.addr_validate(&admin)?, &Empty {})?;

```

hashedone: Isn't it the same just inlined?

jawoznia: It is. This is how it was done earlier so it's just a nitpick. Both versions fine by me

Issue #9, opened 2022-08-15
Improve `dispatch` generation to know what message dispatch to basing on the top level field.


hashedone:
Follow up of #7 

The messages in the contract build with Sylvia are assumed to be in the form of:

```json
{
  "message_type": { ... }
}
```

The `message_type` is determining what message is actually sent. Every trait being `#[interface]` for Sylvia generates structures which are able to deserialize its collection of `message_type`s.

However, for now, generated `dispatch` function is trying to just deserialize the whole message to one of the sub-messages ones by one which might be an unnecessary overhead. Having `const fn messages() -> [&'static str; N]` function for every message type, it is possible  to do the following in the `dispatch` function:
1. Extract the "message_type" from the message send
2. Check which `message()` function returns the array containing "message_type" - as messages returned by `message()` are already sorted, it can be done quickly with binary search
3. Try to deserialize only to the final message type - obviously, it still can fail (as the message may be malformed)

The easy way is to deserialize to `Value`, then check on "message_type" and proceed with deserialization, but I think that it should be possible with custom hand-made `Deserialize` implementation for even more overhead reduction - it is to be verified and tried.

jawoznia:
Delivered under #44

PR #44, opened 2022-09-14
Improving deserialization.

jawoznia:
Instead of attempting to deserialize msg to each possible variant sylvia will try to find which trait implement received message and only then try to deserialize into it.

### DIFF
--- a/sylvia-derive/src/message.rs
+++ b/sylvia-derive/src/message.rs
@@ -662,31 +662,19 @@ impl<'a> GlueMessage<'a> {
 
         let dispatch_arm = quote! {#contract_name :: #contract (msg) =>msg.dispatch(contract, ctx)};
 
+        let recv_msg_name = Ident::new("recv_msg_name", Span::mixed_site());
+
         let deserialization_attempts = interfaces.iter().map(|interface| {
             let ContractMessageAttr {
                 module, variant, ..
             } = interface;
             quote! {
                 let msgs = &#module :: #name ::messages();
-                let map = match val.clone() {
-                    #sylvia ::serde_value::Value::Map(map) => map,
-                    _ => panic!("Unexpected msg format!")
-                };
-                if map.len() != 1 {
-                    panic!("Unexpected number of elements received to deserialize!")
-                }
-                let k = match map.into_iter().next() {
-                    Some(k) => k,
-                    None => panic!("Couldn't access the only element of map!")
-                };
-                if let #sylvia ::serde_value::Value::String(k) = k.0 {
-                    if msgs.into_iter().any(|msg| msg == &k) {
-                        match val.clone().deserialize_into() {
+                if let #sylvia ::serde_value::Value::String(#recv_msg_name) = &#recv_msg_name .0 {
+                    if msgs.into_iter().any(|msg| msg == &#recv_msg_name) {
+                        match val.deserialize_into() {
                             Ok(msg) => return Ok(Self:: #variant (msg)),
-                            Err(err) => return Err(D::Error::custom(format!(
-                                    "Deserialization failed for {}\n{}",
-                                    stringify!(k), stringify!(err)
-                                )))
+                            Err(err) => return Err(D::Error::custom(err)).map(Self:: #variant),
                         };
                     }
                 }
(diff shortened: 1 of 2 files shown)


sylvia-derive/src/message.rs
```diff
@@ -663,35 +663,62 @@ impl<'a> GlueMessage<'a> {
...
+                    _ => panic!("Unexpected msg format!")
+                };
+                if map.len() != 1 {
+                    panic!("Unexpected number of elements received to deserialize!")
+                }
+                let k = match map.into_iter().next() {
+                    Some(k) => k,
+                    None => panic!("Couldn't access the only element of map!")
                 };
+                if let #sylvia ::serde_value::Value::String(k) = k.0 {
```
hashedone, line 682: I am pretty sure all of this can be done once per deserialization, not per attempt. It would also avoid cloning the value all the time. If the item is not a string we just return an error.

sylvia-derive/src/message.rs
```diff
@@ -663,35 +663,62 @@ impl<'a> GlueMessage<'a> {
...
+                module, variant, ..
+            } = interface;
             quote! {
-                let #var = match val.clone().deserialize_into() {
-                    Ok(msg) => return Ok(Self:: #variant (msg)),
-                    Err(err) => err,
+                let msgs = &#module :: #name ::messages();
+                let map = match val.clone() {
+                    #sylvia ::serde_value::Value::Map(map) => map,
+                    _ => panic!("Unexpected msg format!")
```
hashedone, line 673: I guess this message could improve.

sylvia-derive/src/message.rs
```diff
@@ -663,35 +663,62 @@ impl<'a> GlueMessage<'a> {
...
+                };
+                if map.len() != 1 {
+                    panic!("Unexpected number of elements received to deserialize!")
```
hashedone, line 676: And this one.

sylvia-derive/src/message.rs
```diff
@@ -663,35 +663,62 @@ impl<'a> GlueMessage<'a> {
...
+                    None => panic!("Couldn't access the only element of map!")
                 };
+                if let #sylvia ::serde_value::Value::String(k) = k.0 {
+                    if msgs.into_iter().any(|msg| msg == &k) {
+                        match val.clone().deserialize_into() {
+                            Ok(msg) => return Ok(Self:: #variant (msg)),
+                            Err(err) => return Err(D::Error::custom(format!(
+                                    "Deserialization failed for {}\n{}",
+                                    stringify!(k), stringify!(err)
+                                )))
```
hashedone, line 689: Here I would literally just forward an error. We do not need to add more context. So here just `val.deserialize_into().map(Self:: #variant)` and be happy.

sylvia-derive/src/message.rs
```diff
@@ -663,35 +663,62 @@ impl<'a> GlueMessage<'a> {
...
             };
+            if let #sylvia ::serde_value::Value::String(k) = k.0 {
+                if msgs.into_iter().any(|msg| msg == &k) {
+                    match val.clone().deserialize_into() {
+                        Ok(msg) => return Ok(Self:: #contract (msg)),
+                        Err(err) => return Err(D::Error::custom(format!(
+                                "Deserialization failed for {}\n{}",
+                                stringify!(k), stringify!(err)
+                            )))
+                    };
```
hashedone, line 717: Same here, just return original error. You are adding meaningless context, but remove typization.

sylvia-derive/src/message.rs
```diff
@@ -729,16 +756,15 @@ impl<'a> GlueMessage<'a> {
...
                     use serde::de::Error;
+
                     let val = #sylvia ::serde_value::Value::deserialize(deserializer)?;
 
                     #(#deserialization_attempts)*
                     #contract_deserialization_attempt
 
-                    Err(D::Error::custom(format!(
-                        "Expected any of {} or {} messages, but cannot deserialize to neither of those\n{}",
+                    Err(D::Error::custom(format!("Expected any of {} or {} messages, but neither of these was provided.",
```
hashedone, line 765: As I understand properly, you would print here like something like:

> Expected any of Cw1 or ExecMsg messages, but neither of those was provided.

Not very helpful (especially `ExecMsg` part). Better would be:

> Unsupported message received. Messages supported by this contract: `modify_admins`, `freeze`, `execute`.

We have all messages here, so just collect them with [`concat`](https://doc.rust-lang.org/std/primitive.slice.html#method.concat) to a single vector, and then cal [`join`](https://doc.rust-lang.org/std/primitive.slice.html#method.join) on it. It is an additional allocation, but it is an error case - a way less relevant one (better error messages > avoiding every allocation possible). And even this allocation can be avoided (it is preferred here as it is generated so a bit of noise does not do harm): make an array `messages()` results, and do something like:
```rust

let mut msg = all_messages
  .into_iter()
  .flat_map(|messages| messages.iter())
  .fold("Unsupported message received. Messages supported by this contract: ".to_owned(), |mut acc, message|  {
    acc.push_str(message); 
    acc.push_str(", ");
  });
  msg.truncate(msg.len() - 2);
```

Possibly with some special care if there are no messages at all.

hashedone (commented):
We could improve on error handling a bit.

jawoznia:
close #9

hashedone (commented):
Comments for code are small, but it seems to not print errors properly. I added quick temporal tests sending an invalid messages in `subkeys` contract (find it on `improved-dispatch-fail` branch). I got an error message:

> Error parsing into type cw1_subkeys::contract::ContractExecMsg: Unsupported message received. Messages supported by this contract: decrease_allowance, increase_allowance, set_permissions', contracts/cw1-subkeys/src/multitest/tests.rs:51:10

The problem is it shows only messages defined directly on the contract, completely ignoring messages defined on `cw1` and `whitelist` traits (`execute`, `freeze`, and `update_admins`).

Also: rebase to main.

jawoznia:
It seems that on main cw1 messages are not printed either. Possibly it has sth to do with how cw1subkey is defined.

jawoznia:
`Cw1SubkeysContract `implementation was missing:
```
#[messages(cw1 as Cw1)]
#[messages(whitelist as Whitelist)]
```
Now it prints all expected messages.

```
Caused by:
    Error parsing into type cw1_subkeys::contract::ContractExecMsg: Unsupported message received. Messages supported by this contract: execute, freeze, update_admins, decrease_allowance, increase_allowance, set_permissions', contracts/cw1-subkeys/src/multitest/tests.rs:51:10
```

hashedone (approved):
LGTM, Nice work.

PR #48, opened 2022-09-21
Better schema generation

### LINKED ISSUE #18: Figure out what should be generated for better schema generation
@uint is working on a better way of generating schema. Talk to him about details of what to generate.

jawoznia:
close #18


hashedone:
Remember to rebase to main branch

.github/workflows/ci.yml
```diff
@@ -60,6 +60,12 @@ jobs:
         run: cargo install cosmwasm-check
       - name: Check contracts
         run: find target/wasm32-unknown-unknown/release/ -type f -name "*.wasm" -exec cosmwasm-check {} \;
+      - name: Cw1-whitelist schema
+        working-directory: ../contracts/cw1-whitelist/
+        run: cargo schema
+      - name: Cw1-subkeys schema
```
hashedone, line 66: If you are generating schemas it would be probably nice to upload them as artifacts (https://docs.github.com/en/actions/using-workflows/storing-workflow-data-as-artifacts)

contracts/cw1-whitelist/.cargo/config
```diff
@@ -0,0 +1,6 @@
+[alias]
+wasm = "build --release --target wasm32-unknown-unknown"
+wasm-debug = "build --target wasm32-unknown-unknown"
```
hashedone, line 3: Both `wasm` build targets should be performed with `--lib` flag, you will not be able to build schema for wasm. I assume this might be why CI is failing.

contracts/cw1-whitelist/src/contract.rs
```diff
@@ -271,7 +270,7 @@ mod tests {
 
         #[test]
         fn freeze() {
-            let original = whitelist::ExecMsg::Freeze {};
+            let original = whitelist::WhitelistExecMsg::Freeze {};
```
hashedone, line 273: Why? We don't want to prefix message names - it is doubling information and makes it clunky. To avoid name collisions do it only for the glue message, but in a private namespace so it's hidden from the world. The user should use just `whitelis::ExecMsg`.

Unless the bug in the schemars is so crappy that it depends on original message name?

sylvia-derive/src/message.rs
```diff
@@ -247,7 +249,7 @@ impl<'a> EnumMessage<'a> {
                     };
 
                     if attr == ty {
-                        Some(MsgVariant::new(&method.sig, &mut generics_checker))
+                        Some(MsgVariant::new(&method.sig, &mut generics_checker, name))
```
hashedone, line 252: Naaaaaaah, you don't pass the name - the message still should be named as it was. Only an alias for gluing.

hashedone (commented):
The idea is fine, but there is no point to rename those messages where they are generated. You need unique names only for glue messages to overcome the schemars bug, you can alias them in place - it would be better as better names would be generated.

contracts/cw1-whitelist/src/contract.rs
jawoznia: It seems that this bug might be soon fixed though

https://github.com/GREsau/schemars/pull/178

hashedone (approved):
Looks good

PR #49, opened 2022-09-21
Coverage calculation

hashedone:


hashedone:
This is draft. We need [this](https://github.com/CosmWasm/runtime-macros-derive/pull/1) to be merged first, as we want to use `master` git branch for dependency.

jawoznia (approved):
I think it would be nice to make sure that macro expansion is working and that coverage increases with it.

To consider: we could deliver separately change in workflows enabling code coverage and then in second review push runtime-macro-derive and make sure that everything is working fine.

hashedone:
@jawoznia yes, coverage measurement works but a little bit differently than you could expect.

There are three coverages you can consider here:
1. The test coverage - it shows you what is coverage of your test code. Basically, if you create a contract in the test, you measure what is the coverage of your custom contract code. So it is telling you how many of the features used in this test case are actually tested.
2. The generated code coverage - how many of the code generation cases are covered in tests? 100% means, that all code which would be generated is generated somewhere in tests (but it still may be not tested at all). That means that this coverage on its own is useless - it has to be paired with 1). And we want to be as close to 100% as physically possible. The actual coverage could be probably estimated as 2) * 1).
3. Contracts coverage, being just examples coverage - nice to improve here, but not critical. We probably won't bother unless we provide MT utils generation.

PR #51, opened 2022-10-10
Migrate and reply support in sylvia

### LINKED ISSUE #17: Add a `migrate` and `reply` entry points
Add generating the missing `migrate` and `reply` entry point defined on the contract.

It would be nice to have some additional layer on top of it to generate easier differential migrations, and also to connect replies to messages in some way, but for now we want to be able to define those directly.

jawoznia:
close #17


contracts/cw1-subkeys/src/contract.rs
```diff
@@ -274,6 +275,26 @@ impl Cw1SubkeysContract<'_> {
         })
     }
 
+    #[msg(migrate)]
+    fn migrate_0_1_to_0_2(
```
hashedone, line 279: I will name this just migrate. In migrate, one can then dispatch between migrations for specific versions (to we may later do it automagically I think)

contracts/cw1-subkeys/src/multitest/admin_vote_contracts.rs
```diff
@@ -0,0 +1,295 @@
+use anyhow::{bail, Result as AnyResult};
```
hashedone, line 1: Why such a chuge contract in the multitest? Can't you minimize it? Or extract as separate contract?

contracts/cw1-subkeys/src/multitest/admin_vote_contracts.rs
```diff
@@ -0,0 +1,295 @@
...
+            .range(deps.storage, None, None, cosmwasm_std::Order::Ascending)
+            .filter_map(|admin| admin.ok())
+            .map(|(addr, _)| addr)
+            .collect();
+        let resp = AdminsListResp { admins };
+        Ok(resp)
+    }
+
+    #[msg(reply)]
+    pub fn vote_instantiated(&self, ctx: (DepsMut, Env), msg: SubMsgResult) -> StdResult<Response> {
```
hashedone, line 106: I assume you wanted to test the reply, and that is why you put this contract. But contracts are more like examples, not tests. We want to test in `sylvia/tests` where you put a fake contract testing particular utilities. This contract is ok, but extract it as a separate contract, and test it there (it can easily depend on cw1).

Also - because you don't do any smart dispatching, this function should be called `reply` as you handle any replies here, not only `vote_instantiated` (which you expect, but technically you need to verify). Also - I don't see reply id verification. I know that you expect only one reply here, but it is bad habit to assume it, as sometimes it might change in the future.

hashedone (commented):
Nice, you added some examples, but the more important test in `sylvia/tests` - we want there to have very minimalistic contracts test (like contracts having only `migrate` functionality doing something observable, and then making sure that this observable thing happened - same for `reply`).

Some additional comments about code structure, but overall is ok.

sylvia/tests/messages_generation.rs
```diff
@@ -30,27 +30,111 @@ pub trait Interface {
...
+        coef: Decimal::percent(10),
+        desc: "Some description".to_owned(),
+    };
+    let no_args_query = interface::QueryMsg::NoArgsQuery {};
+    let _argumented_query = interface::QueryMsg::ArgumentedQuery {
+        user: Addr::unchecked("owner"),
+    };
+
+    // Ensure no extra variants are generated
+    match no_args_exec {
```
hashedone, line 95: Nice approach, big plus you figured this out

hashedone (approved):
Nice test. I really like this iteration

PR #54, opened 2022-10-17
Increase coverage

### LINKED ISSUE #50: Bring the coverage to reasonable level and add `--fail-under` flag to tarpaulin
Right now we have a way too low coverage - we want to have most of the code covered, I assume ~80-90% (we cannot aim for 100% as tarpaulin gives false negatives, like macro interiors or methods chains). Then we can add a threshold to fail CI with no coverage.

jawoznia:
close #50

Added some tests for contracts.
In case of sylvia implemntation mostly added #[cfg(not(tarpaulin_include))] to ignore false negatives and entry_points implemntation coverage check.

Coverage is still low due to false negatives of code in closures and in quote!.

### DIFF
--- a/contracts/cw1-subkeys/src/lib.rs
+++ b/contracts/cw1-subkeys/src/lib.rs
@@ -7,6 +7,7 @@ pub mod responses;
 pub mod state;
 mod whitelist;
 
+#[cfg(not(tarpaulin_include))]
 #[cfg(not(feature = "library"))]
 mod entry_points {
     use cosmwasm_std::{entry_point, Binary, Deps, DepsMut, Env, MessageInfo, Response};
--- a/contracts/cw1-subkeys/src/lib.rs
+++ b/contracts/cw1-subkeys/src/lib.rs
@@ -7,8 +7,7 @@ pub mod responses;
 pub mod state;
 mod whitelist;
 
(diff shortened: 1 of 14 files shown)


contracts/cw1-subkeys/src/lib.rs
```diff
@@ -7,6 +7,7 @@ pub mod responses;
 pub mod state;
 mod whitelist;
 
+#[cfg(not(tarpaulin_include))]
```
hashedone, line 10: A bit of nitpicking, but I think it is better to collapse those cfgs:
```rust
#[cfg(not(any(feature = "library, tarpaulin_include)))]
```

sylvia-derive/src/parser.rs
```diff
@@ -191,6 +195,7 @@ impl MsgAttr {
         let mut name = Ident::new("InstantiateMsg", content.span());
         let p: Option<Token![,]> = content.parse()?;
 
+        // In what scenario is this true?
```
hashedone, line 198: Not a valid comment. It may be true if some args (I think the name in particular) are passed to instantiate. However I don't know it this makes sense, for now, possibly this chunk could be removed).

hashedone (approved):
LGTM - with some minor comments

PR #67, opened 2022-10-27
Rename args in dispatch to avoid name collission

jawoznia:
Rename args in dispatch.

```
    pub fn dispatch(
        self,
        contract: &Cw1SubkeysContract,
        ctx: (
            cosmwasm_std::DepsMut,
            cosmwasm_std::Env,
            cosmwasm_std::MessageInfo,
        ),
    ) -> std::result::Result<sylvia::cw_std::Response, ContractError> {
        use ExecMsg::*;
        match self {
            IncreaseAllowance {
                spender: field1,
                amount: field2,
                expires: field3,
            } => contract
                .increase_allowance(ctx.into(), field1, field2, field3)
                .map_err(Into::into),
    }
```

### DIFF
--- a/sylvia-derive/src/message.rs
+++ b/sylvia-derive/src/message.rs
@@ -595,8 +595,16 @@ impl<'a> MsgVariant<'a> {
             function_name,
             ..
         } = self;
-        let args = fields.iter().map(|field| field.name);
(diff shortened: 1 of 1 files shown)


hashedone (approved):
LGTM

Issue #53, opened 2022-10-17
Port cw20 contract


hashedone:
[cw20](https://github.com/CosmWasm/cw-plus/tree/main/contracts/cw20-base) is one of the most important contracts in CW ecosystem - we should prioritize migrating it.

PR #68, opened 2022-10-28
Cw20 interfaces

jawoznia:
All interfaces to cw20 beside allowances.

Part of #53


jawoznia:
This will be delivered under different PR

PR #69, opened 2022-11-02
Cw20 contract

jawoznia:
Cw20 allowance and base implementation with coverage of the first one


contracts/cw20-base/src/allowances.rs
```diff
@@ -0,0 +1,359 @@
...
+        let update_fn = |allow: Option<AllowanceResponse>| -> Result<_, _> {
+            let mut val = allow.unwrap_or_default();
+            if let Some(exp) = expires {
+                if exp.is_expired(&env.block) {
+                    return Err(ContractError::InvalidExpiration {});
+                }
+                val.expires = exp;
+            }
+            val.allowance += amount;
+            Ok(val)
```
hashedone, line 77: Why do you force my brain to follow how is `val` change, instead of just:

```rust
match expires {
  let allow = unwrap_or_default();
  let allowance = allow.allowance + amount;
  Some(expires) if !exp.is_expired(&env.block) => Ok(AllowanceResponse {
    allowance,
    expires,
  }),
  None => Ok(AllownaceResponse {
    allowance,
    ..allow
  }),
  _ => Err(ContractError::InvalidExpiration {}),
}
```

contracts/cw20-base/src/allowances.rs
```diff
@@ -0,0 +1,359 @@
...
+        let spender_addr = deps.api.addr_validate(&spender)?;
+        if spender_addr == info.sender {
+            return Err(ContractError::CannotSetOwnAccount {});
+        }
+
+        let key = (&info.sender, &spender_addr);
+
+        fn reverse<'a>(t: (&'a Addr, &'a Addr)) -> (&'a Addr, &'a Addr) {
+            (t.1, t.0)
+        }
```
hashedone, line 111: So what is the point of this function instead of writing `|(a, b)| (b, a)` inline?

contracts/cw20-base/src/contract.rs
```diff
@@ -0,0 +1,280 @@
...
+        Ok(total_supply)
+    }
+
+    #[msg(instantiate)]
+    pub fn instantiate(
+        &self,
+        ctx: (DepsMut, Env, MessageInfo),
+        data: InstantiateMsgData,
+    ) -> Result<Response, ContractError> {
+        let (mut deps, ..) = ctx;
```
hashedone, line 68: I would personally just do `let mut deps = ctx.deps;` here, but full reconstruction is also fine. For more than one field I will definitely deconstruct.

contracts/cw20-base/src/contract.rs
```diff
@@ -0,0 +1,280 @@
...
+        let InstantiateMsgData {
+            name,
+            symbol,
+            decimals,
+            initial_balances,
+            mint,
+            marketing,
+        } = data;
```
hashedone, line 76: Is it worth doing such massive deconstruction? Doing `data.xxx` seems not so much overhead, and this deconstruction looks just big.

contracts/cw20-base/src/contract.rs
```diff
@@ -0,0 +1,280 @@
...
+        // check valid token info
+        validate_msg(&name, &symbol, decimals)?;
+        // create initial accounts
+        let total_supply = Cw20Base::create_accounts(&mut deps, &initial_balances)?;
+
+        if let Some(limit) = mint.as_ref().and_then(|v| v.cap) {
+            if total_supply > limit {
+                return Err(StdError::generic_err("Initial supply greater than cap").into());
+            }
+        }
```
hashedone, line 88: You can collapse it to single if:

```rust
if matches!(
  Some(limit) if total_supply > limit,
  mint.as_ref().and_then(|v| v.cap)
) {
  return Err(...);
}
```

Also this is a perfect place for [`ensure`](https://docs.rs/cosmwasm-std/1.1.5/cosmwasm_std/macro.ensure.html):

```rust
ensure!(
  matches!(
    Some(limit) if total_supply > limit,
    mint.as_ref().and_then(|v| v.cap)
  ),
  Err(...)
);
```

contracts/cw20-base/src/contract.rs
```diff
@@ -0,0 +1,280 @@
...
+    /// Burn is a base message to destroy tokens forever
+    #[msg(exec)]
+    fn burn(
+        &self,
+        ctx: (DepsMut, Env, MessageInfo),
+        amount: Uint128,
+    ) -> Result<Response, ContractError> {
+        let (deps, _, info) = ctx;
+
+        if amount == Uint128::zero() {
```
hashedone, line 182: Another good place for `ensure`.

contracts/cw20-base/src/contract.rs
```diff
@@ -0,0 +1,280 @@
...
+            .add_attribute("from", &info.sender)
+            .add_attribute("to", &contract)
+            .add_attribute("amount", amount)
+            .add_message(
+                Cw20ReceiveMsg {
+                    sender: info.sender.into(),
+                    amount,
+                    msg,
+                }
+                .into_cosmos_msg(contract)?,
```
hashedone, line 250: I would probably extract it to own variable to improve readability a bit

contracts/cw20-base/src/multitest/receiver_contract.rs
```diff
@@ -0,0 +1,139 @@
...
+use cosmwasm_std::{
+    from_slice, Addr, DepsMut, Empty, Env, MessageInfo, Response, StdError, StdResult,
+};
+use cw_multi_test::{App, Contract, Executor};
+use sylvia::contract;
+use thiserror::Error;
+
+use super::receiver::{self, Receiver};
+
+pub struct ReceiverContract {}
```
hashedone, line 11: This should not be empty - all storage atoms should be part of this.

contracts/cw20-base/src/error.rs
```diff
@@ -0,0 +1,41 @@
...
+
+    #[error("Invalid png header")]
+    InvalidPngHeader {},
+
+    #[error("Invalid expiration value")]
+    InvalidExpiration {},
+
+    #[error("Duplicate initial balance addresses")]
+    DuplicateInitialBalanceAddresses {},
+}
```
hashedone, line 41: Get rid of all `{}` in error. They are needed in serde-types to properly (de)serialize them, but in error type, they are completely obsolete and nasty.

hashedone (commented):
Some comments - part of them serious, part minor. Didn't review tests yet.

contracts/cw20-base/src/allowances.rs
```diff
@@ -0,0 +1,359 @@
...
+const MAX_LIMIT: u32 = 30;
+const DEFAULT_LIMIT: u32 = 10;
+
+// this can be used to update a lower allowance - call bucket.update with proper keys
+pub fn deduct_allowance(
+    storage: &mut dyn Storage,
+    owner: &Addr,
+    spender: &Addr,
+    block: &BlockInfo,
+    amount: Uint128,
```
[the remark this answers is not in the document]
jawoznia: as discussed offline although serde-json-wasm should support u128 **_"Error parsing into type cw20_base::contract::ContractExecMsg: u128 is not supported"_** is being returned while processing WasmMsg

contracts/cw20-base/src/contract.rs
jawoznia: there is a lot of usage of the data and in cases of instantiation of TokenInfo it just make it nicer to do 

```
let data = TokenInfo {
            name,
            symbol,
            decimals,
            total_supply,
            mint,
        };
```

then

```
let data = TokenInfo {
            name: data.name,
            symbol: data.symbol,
            decimals: data.decimals
            total_supply,
            mint,
        };
```

But if you prefer it to be a direct usage from data I can change

contracts/cw20-base/src/contract.rs
jawoznia: I decided to do it this way because ctx is a tuple and this was nicer for me then ctx.0;

In the future when we will have struct created to handle context I would also go for let mut deps = ctx.deps

contracts/cw20-base/src/allowances.rs
```diff
@@ -0,0 +1,359 @@
...
+                allowance.expires = exp;
+            }
+            ALLOWANCES.save(deps.storage, key, &allowance)?;
+            ALLOWANCES_SPENDER.save(deps.storage, reverse(key), &allowance)?;
+        } else {
+            ALLOWANCES.remove(deps.storage, key);
+            ALLOWANCES_SPENDER.remove(deps.storage, reverse(key));
+        }
+
+        let res = Response::new().add_attributes(vec![
```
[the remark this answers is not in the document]
jawoznia: I changed it in most cases. Left add_attributes in single place as there would be unnecessary copy of sender f.e. We could avoid it by adding msg as last param though. What do you think?

contracts/cw20-base/src/allowances.rs
```diff
@@ -0,0 +1,332 @@
...
+            |balance: Option<Uint128>| -> StdResult<_> { Ok(balance.unwrap_or_default() + amount) },
+        )?;
+
+        let attrs = vec![
+            attr("action", "send_from"),
+            attr("from", &owner),
+            attr("to", &contract),
+            attr("by", &info.sender),
+            attr("amount", amount),
+        ];
```
hashedone, line 204: Still inconsistent attributes handling. And even if you have reason to do so this way (I assume you just want to create sense attribute as you will consume sender right after - I am fine with it), then at least remove `vec!` as it can be a normal array.

contracts/cw20-base/src/contract.rs
hashedone: It is ok. It was a completely minor note about my preferred style choice, you don't have to make it my way, just consider. If you like this better, go this way.

contracts/cw20-base/src/contract.rs
hashedone: Yea, forwarding fields straightly is a good reason for this deconstruction.

hashedone (commented):
Better, but some majors from previous iteration still not applied.

contracts/cw20-base/src/multitest/receiver_contract.rs
jawoznia: for this contract there is no storage. I can rename it to ReceiverDummyContract maybe to show it is only for testing purposes

contracts/cw20-base/src/allowances.rs
```diff
@@ -0,0 +1,331 @@
...
+    fn decrease_allowance(
+        &self,
+        ctx: (DepsMut, Env, MessageInfo),
+        spender: String,
+        amount: Uint128,
+        expires: Option<Expiration>,
+    ) -> Result<Response, Self::Error> {
+        let (deps, env, info) = ctx;
+
+        let spender_addr = deps.api.addr_validate(&spender)?;
```
hashedone, line 74: Minor, but you don't have to validate this address. Validation takes time (and gas), but if the address is not valid, it would for sure fail on loading allowance (as there is no way an invalid address is inserted into the map). We have `Addr::unchecked(&spender)` for such cases.

contracts/cw20-base/src/contract.rs
```diff
@@ -0,0 +1,337 @@
...
+}
+
+pub struct Cw20Base<'a> {
+    pub(crate) token_info: Item<'static, TokenInfo>,
+    pub(crate) marketing_info: Item<'static, MarketingInfoResponse>,
+    pub(crate) logo: Item<'static, Logo>,
+    pub(crate) balances: Map<'static, &'a Addr, Uint128>,
+    pub(crate) allowances: Map<'static, (&'a Addr, &'a Addr), AllowanceResponse>,
+    // TODO: After https://github.com/CosmWasm/cw-plus/issues/670 is implemented, replace this with a `MultiIndex` over `ALLOWANCES`
+    pub(crate) allowances_spender: Map<'static, (&'a Addr, &'a Addr), AllowanceResponse>,
```
hashedone, line 62: There is a good question if there should be `pub` or `pub(crate)` to make it most useful. There are good arguments for both. For now, leave it as it is, but we need to figure out the proper style in future having it in mind.

hashedone (approved):
I really like it. One minor comment - you can apply the change, but you can leave it as it is; up to you.

PR #70, opened 2022-11-04
Allow nested return types in Queries

jawoznia:
F.e. Option<QueryResponse> is now supported for Query return type


sylvia-derive/src/message.rs
```diff
@@ -518,54 +541,45 @@ impl<'a> MsgVariant<'a> {
...
+        assert_eq!(segments.len(), 1);
+        let segment = &segments[0];
+
+        // In case of aliased result user need to define the return type by hand
+        if segment.ident != "Result" && segment.ident != "StdResult" {
+            panic!(
+                "Neither Result nor StdResult found in return type. \
+                    You might be using aliased return type. \
+                    Please use #[msg(return_type=<your_return_type>)]"
+            );
```
hashedone, line 562: Don't panic to error report. It works to some extend, but better use https://docs.rs/proc-macro-error/latest/proc_macro_error/macro.emit_error.html - it allows you to provide proper span, and in general improves how you can report an error.

hashedone (approved):
Mostly ok, but error could be handled better - please fix before merge

PR #72, opened 2022-11-17
Generate Migrate as struct instead of enum

jawoznia:


sylvia-derive/src/message.rs
```diff
@@ -106,15 +106,19 @@ impl<'a> StructMessage<'a> {
         let (method, msg_attr) = if let Some(method) = methods.next() {
             method
         } else {
-            emit_error!(source.span(), "No instantiation message");
+            if ty == MsgType::Instantiate {
```
hashedone, line 109: Not sure why you check for instantiate too? We want this to check also on migrations, don't we?

hashedone (approved):
LGTM, but not sure about one comment

Issue #27, opened 2022-08-26
Generate Multitest helpers


hashedone:
We want to generate helpers for MTs of some sort. The exact form is to be defined.

I was considering many paths to go, and for now, I think the best is to wrap everything in simple structures. Also, I think that keeping `App` in `RefCell<App>` would be beneficial as wrappers could keep the `app` and call mutable functions on it. It hits performance a bit, but it is only for test purposes so I don't think it matters so much. I think about introducing a wrapper like:

```rust
#[derive(Clone)]
struct App {
  app: RefCell<cw_multi_test::App>,
}
```

With all the functions forwarded to the app so it is transparent in usage in multitest, but because it is RC under the hood we can easily pass it to all generated wrappers. 

Then for interfaces, I would just generate accessors like this:

```rust
#[interface]
trait Cw1 {
  #[msg(exec)] 
  fn add_member(&self, ctx: ExecCtx, member: String) -> Result<Response, Self::Error>;

  #[msg(query)]
  fn find_member(&self, ctx: QueryCtx, member: String) -> Result<FindMemberResponse, Self::Error>;
}

// All below to be generated
pub struct Cw1Proxy<'app> {
  contract: Addr,
  app: &'app sylvia::App,
}

impl<'app> Cw1Proxy<'app> {
  fn new(contract: Addr, app: &'app sylvia::App) -> Self { /* ... */ }

  fn add_member(&self, params: impl sylvia::ExecParams, member: String) -> Result<Response, Sef::Error> { /* ... */ }
  
  fn find_member(&self, member: String) -> Result<FindMemberResponse, Self::Error>;
}

pub struct Cw1ExecProxy<'proxy> {
  proxy: &'proxy Cw1Proxy<'proxy>,
  sender: Addr,
  funds: Vec<Coin>,
}
```

The `ExecParams` trait is delivering all additional data for the message to be sent. It always contains the message sender, but may also contain additional things like funds (I don't have anything more than that, but there should be a way to configure any additions to execution). To make it easy to use, it should be implemented on `&Addr`/`&str` (which is sending a message with sender and nothing more) and `(&Addr, &[Coin])`/`(&'str, &[Coin])` for sending a message with funds - but some more explicit types for such (like `WithFunds<'a>(&'a Addr, &'a [Coin])` would also be good to deliver. The idea is to hide unnecessary things which are not sent in most cases.

For contracts, things would look very similar, but a bit more is generated:

```rust
[...]

PR #76, opened 2022-12-21
Multitest helpers

jawoznia:


contracts/cw1/Cargo.toml
```diff
@@ -12,4 +12,8 @@ homepage = "https://cosmwasm.com"
 cosmwasm-std = { version = "1.1", features = ["staking"] }
 cosmwasm-schema = "1.1"
 serde = { version = "1", default-features = false, features = ["derive"] }
-sylvia = { version = "0.2.2", path = "../../sylvia" }
+sylvia = { version = "0.2.2", path = "../../sylvia", features = ["mt"] }
```
hashedone, line 15: We don't want Sylvia to have "mt" in a normal build. Add it to dev-dependencies, and add an additional feature "mt" feature which transitively enables "mt" feature for Sylvia.

hashedone (commented):
Seems mostly ok, but I would like to see the usage of emitted stuff here, so we see it works as we want it to work. Also important comment about deps.

contracts/cw1/Cargo.toml
jawoznia: As for example this is only part of implementation containing helpers for traits.
There will be a second delivery with the multitest helpers generation for contracts. Then there will be examples.

jawoznia:
part of #27

hashedone (approved):
For now it is fine, but in follow up we need examples of usage of that (but I get we need contract stuff generation for that)

PR #88, opened 2023-03-22
Better proxy generation

jawoznia:


jawoznia:
This change aims to make adding #[cfg(test)] use ../..Proxy; obsolete in code.

#[Interface] generates Proxy type

#[contract] impl Trait for Contract generates new Trait with message methods which are implemented on the Proxy type

hashedone (approved):
Besides I think it might be simpler, I think it is good for now - we will check out how it works out, it is blocking you for too long.

PR #89, opened 2023-03-22
Exposing generated multitest_utils via "mt" feature

jawoznia:


sylvia-derive/src/multitest.rs
```diff
@@ -268,18 +268,18 @@ impl<'a> MultitestHelpers<'a> {
             .collect();
 
         quote! {
-            #[cfg(test)]
+            #[cfg(any(test, feature = "mt"))]
```
hashedone, line 271: There should be no cfg attribute generated at all. Generation of the whole chunk should be enabled/disabled based on the state of sylvia feature flag. Here you assume you have any control about feature of crate using your toolbox which is a mistake. Just remove the whole `cfg` and make sure it is not generated when `mt` feature is not enabled on `syvia-derive`.

hashedone (commented):
One little problem commented, besides that it is good.

hashedone (approved):
LGTM

jawoznia:
Closing due to #88 being delivered

PR #90, opened 2023-03-24
Generate entry points

### LINKED ISSUE #38: Entry points generation
Writing entry points is very tedious right now. Create a new contract instance, and call dispatch on it.  Maybe we would be able to generate it somehow.

jawoznia:
close #38

### DIFF
--- a/sylvia-derive/Cargo.toml
+++ b/sylvia-derive/Cargo.toml
@@ -12,7 +12,6 @@ categories = ["api-bindings", "wasm"]
 
 [features]
 mt = []
-no_entry_points = []
 
 [lib]
 proc-macro = true
--- a/sylvia-derive/Cargo.toml
+++ b/sylvia-derive/Cargo.toml
@@ -12,18 +12,19 @@ categories = ["api-bindings", "wasm"]
 
 [features]
 mt = []
+library = []
 
 [lib]
 proc-macro = true
 
 [dependencies]
 syn = { version = "1.0.107", features = [
-  "parsing",
-  "derive",
-  "fold",
-  "visit",
-  "full",
-  "extra-traits",
+    "parsing",
+    "derive",
+    "fold",
+    "visit",
+    "full",
+    "extra-traits",
 ] }
 quote = "1.0.23"
 proc-macro2 = "1.0.50"
--- a/sylvia-derive/Cargo.toml
+++ b/sylvia-derive/Cargo.toml
@@ -12,19 +12,19 @@ categories = ["api-bindings", "wasm"]
 
 [features]
 mt = []
-library = []
+no_entry_points = []
 
 [lib]
 proc-macro = true
 
 [dependencies]
 syn = { version = "1.0.107", features = [
-    "parsing",
-    "derive",
-    "fold",
-    "visit",
-    "full",
-    "extra-traits",
+  "parsing",
+  "derive",
+  "fold",
+  "visit",
+  "full",
+  "extra-traits",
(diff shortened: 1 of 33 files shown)


hashedone (commented):
Wow, I need to taka a closer look, but I won't expect it would be so simple. I would very like to merge this.

jawoznia:
Main problem here is with cargo features. Unfortunately features are unionized in the scope of the workspace and entry points are not generated because "library" feature is set in cw1_subkeys for cw1_whitelist deps.

sylvia-derive/Cargo.toml
```diff
@@ -12,18 +12,19 @@ categories = ["api-bindings", "wasm"]
 
 [features]
 mt = []
+library = []
```
hashedone, line 15: This is not a valid feature name. Sylvia is always used as a library. Rename to `entry_points`.

sylvia-derive/src/input.rs
```diff
@@ -153,9 +154,15 @@ impl<'a> ImplInput<'a> {
             #exec
 
             #query
+
+            #entry_points
```
hashedone, line 158: I don't see where do you disable generation entry points? We cannot generate them when the feature flag for it is not enabled (or the other way).

sylvia-derive/Cargo.toml
jawoznia: because as far as I'm concerned we cannot disable features and also I think by default entry_points should be disabled I will change it to `no_entry_points` and users can define on their side some feature f.e. `library` that if set would also set `no_entry_points` for sylvia

sylvia-derive/src/input.rs
jawoznia: Here https://github.com/CosmWasm/sylvia/pull/90/files#diff-83fed236913cceb2202f2c224d7b2096cd8c557a9b9808153c34f27eb0ca33daR863

hashedone: Ok, I missed this line, its ok

sylvia-derive/Cargo.toml
hashedone: Ya, this is also somehow related to the recent discussion to change it the other way... but doing it as `no_entry_points` is completely fine.

However, I am pretty sure changing it to `entry_points`, and then adding the reversed logic - adding the `binary` feature and enabling it only on the contracts being built as binaries should solve the problem. Unfortunately I am not sure how it would work with workspace optimizer :/

hashedone (commented):
We cannot have to fail CI because of some features issue.

I think the way to fix this is to change the CI flow - instead of running the `cargo test` on the workspace, we need to create a separate job for every contract separately, which calls `cargo test` in the contract directory.

sylvia-derive/Cargo.toml
jawoznia: We could do that but I think from user perspective it would be weird to add "entry_points" flag to disable them. That's why I opt for the less nice but making more sense "no_entry_points"

jawoznia:
> We cannot have to fail CI because of some features issue.
> 
> I think the way to fix this is to change the CI flow - instead of running the `cargo test` on the workspace, we need to create a separate job for every contract separately, which calls `cargo test` in the contract directory.

It won't work as `cw1_subkeys` uses `cw1_whitelist` as a dependency. So running build for `subkeys` will cause in the current state that no `entry points` will be generated

hashedone (approved):
LGTM

PR #91, opened 2023-03-28
Introduce context struct

### LINKED ISSUE #30: Introduce custom context types

jawoznia:
close #30

### DIFF
--- a/contracts/cw1-subkeys/src/contract.rs
+++ b/contracts/cw1-subkeys/src/contract.rs
@@ -48,7 +48,6 @@ impl Cw1SubkeysContract<'_> {
         admins: Vec<String>,
         mutable: bool,
     ) -> Result<Response, ContractError> {
-        // here deps was branched
         let result = self.whitelist.instantiate(
             InstantiateCtx {
                 deps: ctx.deps.branch(),
--- a/contracts/cw1-subkeys/src/contract.rs
+++ b/contracts/cw1-subkeys/src/contract.rs
@@ -1,12 +1,13 @@
 use cosmwasm_std::{
-    ensure, ensure_ne, Addr, BankMsg, Coin, CosmosMsg, Deps, DepsMut, DistributionMsg, Env,
-    MessageInfo, Order, Response, StakingMsg, StdResult,
+    ensure, ensure_ne, Addr, BankMsg, Coin, CosmosMsg, Deps, DistributionMsg, Env, Order, Response,
+    StakingMsg, StdResult,
 };
 use cw1_whitelist::contract::Cw1WhitelistContract;
 use cw1_whitelist::whitelist;
 use cw2::set_contract_version;
 use cw_storage_plus::{Bound, Map};
 use cw_utils::Expiration;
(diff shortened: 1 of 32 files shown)


sylvia-derive/src/parser.rs
```diff
@@ -109,15 +109,20 @@ impl MsgType {
...
+                #sylvia ::types::InstantiateCtx
             },
             Migrate => quote! {
-                (cosmwasm_std::DepsMut, cosmwasm_std::Env)
+                #sylvia ::types::MigrateCtx,
             },
             Query => quote! {
-                (cosmwasm_std::Deps, cosmwasm_std::Env)
+                #sylvia ::types::QueryCtx
             },
```
hashedone, line 126: Here I am not exactly sure - we obviously want to support the new context types, but supporting original tuples or custom context types would also be nice. Iw ould prefer here to emit something like:

```
impl From<(#sylvia ::cw_std::DepsMut, #sylvia ::cw_std::Env, #sylvia ::cw_std::MessageInfo)>
```

I am not sure, but I think we use this in function arguments position, and this way you will be able to use any type constructible from tuple context. Obviously some `.into()` in generated could would have to be added.

hashedone (commented):
We could consider how we emit signatures of functions taking context types.

sylvia-derive/src/parser.rs
jawoznia: reverted to simple tuple

contracts/cw1-subkeys/src/contract.rs
```diff
@@ -43,62 +44,63 @@ impl Cw1SubkeysContract<'_> {
...
-                .unwrap_or_default();
-
-            let mut allowance = allow
-                .filter(|allow| !allow.expires.is_expired(&env.block))
-                .unwrap_or_default();
+        let spender = ctx.deps.api.addr_validate(&spender)?;
+        ensure_ne!(
+            ctx.info.sender,
+            spender,
+            ContractError::CannotSetOwnAccount {}
```
hashedone, line 73: Why use `{}` for errors? They don't need it; they are not serialized.

hashedone (approved):
LGTM

contracts/cw1-subkeys/src/contract.rs
jawoznia: Agree. This is unnecessary

jawoznia: Will deliver fix in separate PR

jawoznia: Removal of curly braces on errors #95

Issue #92, opened 2023-03-29
Migrate tests to use generated Proxy


jawoznia:
See cw1_whitelist tests

PR #93, opened 2023-03-29
cw1-subkeys tests migrated to generated proxy

jawoznia:
Part of #92


contracts/cw1-subkeys/src/multitest/tests.rs
```diff
@@ -31,18 +30,20 @@ macro_rules! assert_sorted_eq {
...
 #[test]
 fn get_contract_version_works() {
-    let mut app = App::default();
+    let app = App::default();
 
-    let owner = Addr::unchecked("owner");
+    let owner = "owner";
 
-    let code_id = Cw1SubkeysCodeId::store_code(&mut app);
+    let code_id = Cw1SubkeysContractCodeId::store_code(&app);
```
hashedone, line 37: I am not a fan of the name `Cw1SubkeysContractCodeId` - why there is this `Conract` generated in the name? It it loooong.

contracts/cw1-subkeys/src/multitest/tests.rs
```diff
@@ -54,168 +55,175 @@ fn get_contract_version_works() {
...
+        let app = App::default();
 
-        let owner = Addr::unchecked("owner");
-        let spender1 = Addr::unchecked("spender1");
-        let spender2 = Addr::unchecked("spender2");
-        let spender3 = Addr::unchecked("spender3");
+        let owner = "owner";
+        let spender1 = "spender1";
+        let spender2 = "spender2";
+        let spender3 = "spender3";
```
hashedone, line 72: Slight suggestion - I would make it an array:
```
let spenders = ["spender1", "spender2", "spender3"];
```

contracts/cw1-subkeys/src/multitest/tests.rs
```diff
@@ -31,18 +30,20 @@ macro_rules! assert_sorted_eq {
...
     let contract = code_id
-        .instantiate(&mut app, &owner, &[], true, "Sublist contract")
+        .instantiate(vec![owner.to_owned()], true)
+        .with_label("Sublist contract")
+        .call(owner)
         .unwrap();
 
     let version: ContractVersion =
-        query_contract_info(&app.wrap(), contract.addr().clone()).unwrap();
+        query_contract_info(&app.app.borrow().wrap(), contract.contract_addr.to_string()).unwrap();
```
hashedone, line 46: The `sylvia::App` should have some of those functions exposed - in particular `wrap()` so it is callable directly.

contracts/cw1-subkeys/src/multitest/tests.rs
```diff
@@ -54,168 +55,175 @@ fn get_contract_version_works() {
...
 
         let contract = code_id
-            .instantiate(&mut app, &owner, &[&owner], false, "Sublist contract")
+            .instantiate(vec![owner.to_owned()], false)
+            .with_label("Sublist contract")
+            .call(owner)
             .unwrap();
 
-        let height = app.block_info().height;
+        let height = app.app.borrow().block_info().height;
```
hashedone, line 129: `block_info()` should also be exposed on the `sylvia::App` directly.

contracts/cw1-subkeys/src/multitest/tests.rs
```diff
@@ -54,168 +55,175 @@ fn get_contract_version_works() {
...
+                spender.to_owned(),
                 coin(1, ATOM),
-                Expiration::AtHeight(height + 1),
+                Some(Expiration::AtHeight(height + 1)),
             )
+            .call(owner)
             .unwrap();
 
-        app.update_block(next_block);
+        app.app.borrow_mut().update_block(next_block);
```
hashedone, line 139: Also to be exposed on `sylvia::App`.

hashedone (commented):
I would change how is the name of `CodeId` generated as it is getting long. Also as you update `App`, all functions which are possible to be exposed on it.

contracts/cw1-subkeys/src/multitest/tests.rs
jawoznia: unfortunately we cannot expose wrap through sylvia::App. borrow().wrap() will try to return reference to local value. We can however hide the borrow behind the app() and app_mut() or something like that

hashedone (approved):
LGTM

PR #101, opened 2023-04-20
Additional attribute for entry points

hashedone:
The whole trick here is moving the entire entry points generation to its own macro.

This allows enabling/disabling it based on some criteria (like feature being enabled) with `cfg_attr`.

Then there is an intermediate problem - `ContractError` has to be somehow passed to the `entry_points` macro. We don't want to duplicate it.

The easiest solution is to move the `ContractError` definition. This allows you to read it by all the macros. The important thing is that the `entry_points` macro does not strip the internal attributes - the fact that attributes are evaluated top-to-bottom is used here. It only reads the attribute.

An additional step to be done is to do the same thing for `ContractError` for interfaces and for the `module` thingy - it would make it more consistent.

### DIFF
--- a/contracts/cw1-subkeys/src/contract.rs
+++ b/contracts/cw1-subkeys/src/contract.rs
@@ -9,6 +9,9 @@ use cw_storage_plus::{Bound, Map};
 use cw_utils::Expiration;
 use sylvia::{contract, schemars};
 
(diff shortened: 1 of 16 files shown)


jawoznia (approved):
LGTM

Issue #102, opened 2023-04-28
Explain use of messages


ethanfrey:
Also that the trait name must be camel case of the package it is in.
We must import the package into the namespace of the contract using it.

Also why we need

```rust
#[contract(error=ContractError)]
#[messages(vault_api as VaultApi)]
impl VaultContract<'_> {
}
```

Issue #103, opened 2023-04-28
Docs on how intermeditate/advanced CosmWasm devs can start using it


ethanfrey:
* Explain basic patterns (traits, interfaces, contracts)
* Show how to view auto-generated types (eq QueryMsg) via `cargo doc`
* Explain what this `#[messages]` thing means
* Show how to query a remote contract (that was defined with traits)
* Show how to use reply block (just what you showed me today)
* Other details...
* Link to roadmap of some soon upcoming fixes

Issue #28, opened 2022-08-26
Generate queriers for interfaces


hashedone:
It would be helpful to provide utilities to easily query the contract.

For interfaces:

```rust
#[interface]
pub trait Cw1 {
  type Error: From<StdError>;
  
  #[msg(query)]
  fn find_member(&self, ctx: QueryCtx, member: String) -> Result<FindMemberResponse, Self::Error>;
}

// Generated:
pub struct Cw1Querier<'a> {
  contract: &'a Addr,
  querier: &'a cosmwasm_std::Querier,
}

impl<'a> Cw1Querier<'a> {
  pub fn new(querier: &'a cosmwasm_std::Querier, contract: Addr) -> Self { /* ... */ }
  
  pub fn find_member(&self, member: String) -> Result<FindMemberResponse, Self::Error>;
}
```

Additionally for contracts:

```rust
#[contract]
#[messages(cw1 as Cw1)]
pub struct Cw1WhitelistContract { /* ... */ }

#[derive(Serialize, Deserialize)]
pub struct Cw1WhitelistContractProxy(Addr);

impl Cw1WhitelistContractProxy {
  pub fn addr(&self) -> &Addr;
  
  pub fn cw1<'a>(&'a self, querier: &'a cosmwasm_std::Querier) -> Cw1Querier<'a> { /* .. */ }
}
```

This would allow keeping the contract in storage with semantical contract type, and then directly call queries. The access to underlying address is obviously required (eg. for sending submessages). Also implementing `impl From<Cw1WhitelistContractProxy> for Addr` should probably happen for free unwrapping.

ethanfrey:
This would be a great help to add.
(Also explanation on how to use cargo doc or other tools to see it)

ethanfrey:
Hey @hashedone I wrote some such structs for use with mesh-security, focusing just on calling the traits / interfaces (which is what we are importing).

Check out https://github.com/osmosis-labs/mesh-security/pull/21 

I think the code there could totally be auto-generated (I added the last `funds: Vec<Coin>` argument to show that). Maybe a bit of cleanup on the API, but no need to add yet. I like adding some contract-specific one and a builder for instantiate and all, but that is another level. For now, ensuring we properly call remote contracts that implement some expected interface is a clear and useful use-case and doesn't seem like too much lift.

Let me know if this looks doable.

hashedone:
[...]

PR #106, opened 2023-05-02
Improved README

hashedone:
There is some bug with multitest generation which makes it invalid right now - figuring it out.


hashedone:
Closes #102 #103

README.md
```diff
@@ -4,37 +4,680 @@ Sylvia is the old name meaning Spirit of The Wood.
...
+
+#[entry_point]
+fn query(deps: Deps, env: Env, msg: ContractQueryMsg) -> Result<Binary, ContractError> {
+    msg.dispatch(&CONTRACT, (deps, env))
+}
+```
+
+## Interfaces
+
+One of the fundamental ideas of Sylvia's framework is interfaces, allowing the
```
jawoznia, line 317: are

README.md
```diff
@@ -4,37 +4,680 @@ Sylvia is the old name meaning Spirit of The Wood.
...
+    let member = "john";
+
+    let code_id = contract::multitest_utils::CodeId::store_code(&app);
+
+    let contract = code_id.instantiate(0)
+        .with_label("My contract")
+        .call(owner);
+
+    contract
+        .group()
```
jawoznia, line 623: group_proxy()

README.md
```diff
@@ -4,37 +4,680 @@ Sylvia is the old name meaning Spirit of The Wood.
...
+
+To use sylvia in the contract, you need to add couple dependencies - sylvia itself,
+and additionally: `cosmwasm-schema`, `schemars` and `cosmwasm_std`.
+
+```shell
+$ cargo add syvia cosmwasm-schema schemars cosmwasm-std
+...
+```
+
+You should also make sure your crate is compiling as `cdylib`, setting the proper
```
ethanfrey, line 51: This is useful instructions.  But two points:

1. I would break "create a new contract" into it's own markdown file (linked from this)
2. People shouldn't do all these steps, but could do something like [cw-template](https://github.com/CosmWasm/cw-template) to generate a basic sylvia contract (feel free to make it simpler / targetted to multi-contract repos)

README.md
```diff
@@ -4,37 +4,680 @@ Sylvia is the old name meaning Spirit of The Wood.
...
+        Ok(Response::new())
+    }
+}
+```
+
+Sylvia would add the field into the instantiation message, which now becomes this:
+
+```rust
+struct InstantiateMsg {
+    counter: 64,
```
ethanfrey, line 189: typo - u64

README.md
```diff
@@ -4,37 +4,680 @@ Sylvia is the old name meaning Spirit of The Wood.
...
+
+```rust
+#[entry_point]
+fn execute(deps: DepsMut, env: Env, info: MessageInfo, msg: ContractExecMsg) -> StdResult<Response> {
+    msg.dispatch(&CONTRACT, (deps, env, info))
+}
+```
+
+One problem you might face now is that we use the `StdResult` for our contract,
+but we often want to define the custom error type for our contracts - hopefully,
```
ethanfrey, line 238: As in https://github.com/CosmWasm/book/pull/39... it's fortunately (not hopefully)

README.md
```diff
@@ -4,37 +4,680 @@ Sylvia is the old name meaning Spirit of The Wood.
...
+#[contract]
+#[messages(group as Group)]
+impl MyContract<'_> {
+    // Nothing changed here
+}
+```
+
+Here are a couple of things to talk about.
+
+First, note that I defined the interface trait in its separate module with a name
```
ethanfrey, line 389: Please explain this a bit more in detail with examples:

`mod group` -> `trait Group`
`mod cross_staking` -> `trait CrossStaking`

README.md
```diff
@@ -4,37 +4,680 @@ Sylvia is the old name meaning Spirit of The Wood.
...
+modules which is sometimes useful.
+
+Going through the doc, you will see that all messages are generated in their structs/traits
+modules. To send messages to the contract, we can just use them:
+
+```rust
+use sylvia::cw_std::{WasmMsg, to_binary};
+
+fn some_handler(my_contract_addr: String) -> StdResult<Response> {
+    let msg = my_contract_crate::Execute::Increment {};
```
ethanfrey, line 435: This isn't `my_contract_crate::MyContractExecMsg::Increment {}` ??

I am seeing stuff like this when coding.

README.md
```diff
@@ -4,37 +4,680 @@ Sylvia is the old name meaning Spirit of The Wood.
...
+`--document-private-item` is optional, but it will generate documentation of not-public
+modules which is sometimes useful.
+
+Going through the doc, you will see that all messages are generated in their structs/traits
+modules. To send messages to the contract, we can just use them:
+
+```rust
+use sylvia::cw_std::{WasmMsg, to_binary};
+
+fn some_handler(my_contract_addr: String) -> StdResult<Response> {
```
ethanfrey, line 434: Let's make this easier with #28 and you can update this code to show how to use the handlers...
Not blocking a merge, but a note for later

README.md
```diff
@@ -4,37 +4,680 @@ Sylvia is the old name meaning Spirit of The Wood.
...
+
+        contract.increment().call(owner).unwrap();
+
+        let counter = contract.counter().unwrap();
+        assert_eq!(counter, contract::CounterResp { counter: 4});
+    }
+}
+```
+
+First of all, note the `contract` module I am using here - it is a slight change
```
ethanfrey, line 564: Very nice explanation of multi-test usage

ethanfrey (commented):
Nice docs and explanation.

Added some minor comments, but generally good to merge when addressed.

The other meta-issue, is this is huge, and maybe best to break into 4-5 different pages and link them from a directory in the README. Keeping the intro explanation, then links to how to use, then the Roadmap. Maybe a few other key points in the README, but keeping the detailed codingg instructions linked from README more like a book.

(I prefer the docs here than in book now, as easier to find, and the code is changing so wil need to be updated with future PRs)

README.md
hashedone: Actually typo - it is `my_contract_crate::ExecMsg::Increment {}`

README.md
```diff
@@ -4,37 +4,680 @@ Sylvia is the old name meaning Spirit of The Wood.
...
+struct InstantiateMsg {
+    counter: 64,
+}
+```
+
+What is essential - the field in the `InstantiateMsg` gets the same name as the
+function argument. Be careful using the typical Rust pattern to prefix it with
+`_` to leave it unused. Unfortunately, to properly silence the unused warning here,
+you need to add the `#[allow(unused)]` attribute before the argument or the whole
+function.
```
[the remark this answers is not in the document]
hashedone: I completely forgot it finally get to be implemented

README.md
hashedone: I really don't like this proxy suffix tbh :/ Just because we named type this way doesn't have to make the call have some suffixed added. Anyway - we will get rid of this call as soon as #109 is done.

README.md
hashedone: Splitting it into different files is definitely the way to go. However API is not stabilized yet, and it would be easier to check one file before release to update it, than the whole set of them.

Cargo template is a good idea. Just created #110. Will be part of 0.5 (I really don't want to add anything to 0.4).

README.md
hashedone: #28 is not yet implemented. It Is README for the state of today. Will be updated before 0.5 release

README.md
hashedone: Unfortunately not true before 0.4 because of #108 has to be first merged.

jawoznia (approved):

PR #108, opened 2023-05-04
Fix for multitests generation

hashedone:
This fixes how MT utils are generated (not only for test passes) and makes names a bit more user-friendly.


jawoznia (approved):
LGTM

PR #115, opened 2023-05-05
Incremental migrations support

jawoznia:


sylvia-derive/src/message.rs
```diff
@@ -14,12 +15,242 @@ use syn::parse::{Parse, Parser};
...
+
+                    let mut resp = #sylvia ::cw_std::Response::new();
+                    let results: Vec< #result > = variants.iter().map(|v| match v {
+                        #(#dispatch_arms),*
+                    }).collect();
+                    for result in results {
+                        let result = result?;
+                        resp = resp.add_events(result.events);
+                        resp = resp.add_attributes(result.attributes);
+                    }
```
jawoznia, line 244: This is the most uncertain part of the PR. Because we call multiple migrate methods returning multiple Results we should merge them somehow not to loose any info. Currently `data: Binary` is not concatenated. Also submsgs and messages are left.

jawoznia:
closing as this PR is outdated

Issue #111, opened 2023-05-05
Generating queriers for Traits


hashedone:
Part of #28

PR #117, opened 2023-05-10
Query helpers

jawoznia:


ethanfrey (commented):
This is a good start.
I would like something closer to what I did in my PR.

Including message generation is very important to me.

Making a serializable object is a nice-to-have, not required for now if that is too hard, but ideal for a near-future release.

hashedone (commented):
As far as I understand, the code query helpers are generated properly, but they are named `InterfaceNameQuerier`, but it is a duplication of `InterfaceName` and confusing - we always want to have a single "message source" (contract on the interface) per module, and we just spawn generated things into the module - here it should be named simply `Querier`.

In the future, we will probably move all the generated sylvia things into an additional `sv` module simply to avoid naming collisions with user-defined types, but we don't do it now as it requires us to do it everywhere.

Also, tests are not great. You just slam the `use super::*` in the contract module, which is a no-go. You can do that for "layering" modules where you just want to separate tests, but not in modules with a particular purpose. Now looking only at the test, I don't really know where the `CounterQuerier` really comes from. If it is from the `counter` module (which I think is the case), then all is fine. If it is from the `super` module, it is bad.

Finally - there are no tests for querier for contract messages (non-trait). I believe they are generated correctly, but there is no test for that.

jawoznia:
close #28 
close #117
close #111

hashedone:
Closing in favor of alternative solution

PR #121, opened 2023-05-17
Remote structure

jawoznia:
PR1 of #28


sylvia/tests/remote.rs
```diff
@@ -0,0 +1,55 @@
...
+
+    impl SomeInterface for super::SomeContract {
+        type Error = StdError;
+    }
+}
+
+#[contract]
+#[messages(some_interface as SomeInterface)]
+impl SomeContract {
+    #[allow(clippy::new_without_default)]
```
hashedone, line 23: Naaaaaaaaaaaaaaaaaaaaaaaaaaaaaaah, this is ZST. Don't even implement new on it, you just create it in place.

sylvia/tests/remote.rs
```diff
@@ -0,0 +1,55 @@
+use cosmwasm_std::{Response, StdResult};
+use sylvia::{contract, types::InstantiateCtx};
+
+pub struct SomeContract {}
```
hashedone, line 4: You ain't need `{}` here. ZST you are doing just `pub struct SomeContract;`.

sylvia/tests/remote.rs
```diff
@@ -0,0 +1,55 @@
...
+}
+
+#[cfg(test)]
+mod tests {
+    use cosmwasm_std::Addr;
+
+    use crate::some_interface;
+
+    #[test]
+    fn remote_generation() {
```
hashedone, line 41: Good.

hashedone (commented):
Overall good. The utility for accessing `Addr` remotely is missing. Additionally, minor fixes in tests.

sylvia/tests/remote.rs
jawoznia: We require the `new` method in both entry_points and multitests generation. It is currently obligatory to implement it on contract.

hashedone: Ok, makes sense - still, remove `{}`, and just derive default as it is oneliner.

jawoznia: We could check if there is a new method and otherwise create the contract like unit type.

jawoznia: or require the default trait implementation on it

sylvia-derive/src/input.rs
```diff
@@ -62,12 +64,17 @@ impl<'a> TraitInput<'a> {
                     #messages
 
                     #multitest_helpers
+
+                    #remote
```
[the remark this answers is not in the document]
jawoznia: Seems skipping tarpaulin didn't do the job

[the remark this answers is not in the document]
jawoznia: Wrapping quote in scope did the job although I don't really like this extra nesting

sylvia/tests/remote.rs
hashedone: No, don't diverge basing on existence of `new()`. This is a bookcase for Default trait

hashedone (approved):
LGTM

PR #122, opened 2023-05-18
Querier generation

jawoznia:
PR2 of #28


sylvia-derive/src/querier.rs
```diff
@@ -0,0 +1,119 @@
...
+                        None
+                    }
+                }
+                _ => None,
+            })
+            .collect();
+        Self { variants }
+    }
+
+    pub fn for_interface(source: &'a ItemTrait, generics: &[&'a GenericParam]) -> Self {
```
hashedone, line 47: `for_contract` and `for_interface` are almost identical - at first glance, they only differ on the `source` type, but then just reach very similar `items` under the hood, doing exactly the same things on them.

That means that you want to abstract type differences. Here you basically reach the `Method`  variant, very similar for both, only to access common fields `sig` and `attrs` which are identical at this point.

What you do to make it properly is either:
1. Implement extension trait on `TraitItemMethod` and `ImplItemMethod`, which gives you access to `attrs` and `sig`, and then extension trait for `ItemTrait` and `ImplTrait` to give you back iterator over something implementing your trait - it involves generic traits and possibly GATs, it is a difficult approach, most possibly overkill.
2. Map `ItemTrait` and `ItemImpl` to some iterator over temporary struct before, and pass this iterator into this function argument. This looks like the best approach. Ideally you create a trait on both types with a function returning iterator over your simplified `items` containing only signatures and attributes of whatever methods you found.

sylvia-derive/src/querier.rs
```diff
@@ -0,0 +1,119 @@
...
+            let parameters = fields.iter().map(MsgField::emit_without_attrs);
+            let fields_names = fields.iter().map(MsgField::name);
+            let variant_name = Ident::new(&name.to_string().to_case(Case::Snake), name.span());
+
+            quote! {
+                fn #variant_name(&self, #(#parameters),*) -> Result< #return_type, #sylvia:: cw_std::StdError> {
+                    let query = QueryMsg:: #variant_name (#(#fields_names),*);
+                    self.querier.query_wasm_smart(self.contract, &query)
+                }
+            }
```
hashedone, line 91: This whole thing belongs to `MessageVariant` - this way you don't need to make its fields public.

sylvia-derive/src/querier.rs
```diff
@@ -0,0 +1,119 @@
...
+use crate::check_generics::CheckGenerics;
+use crate::crate_module;
+use crate::message::{MsgField, MsgVariant};
+use crate::parser::{MsgAttr, MsgType};
+
+pub struct Querier<'a> {
+    variants: Vec<MsgVariant<'a>>,
+}
+
+impl<'a> Querier<'a> {
```
hashedone, line 18: In general - I am pretty sure almost the whole of this thing belongs to the message variant, which is built first from the `ItemTrait` or `ItemImpl`, and then just some `emit_querier` is called on it (more or less, probably something more on that which comes out while implementing).

sylvia/tests/querier.rs
```diff
@@ -0,0 +1,125 @@
...
+
+#[derive(
+    Serialize, Deserialize, Clone, PartialEq, Eq, sylvia::schemars::JsonSchema, Debug, Default,
+)]
+pub struct CountResponse {
+    pub count: u64,
+}
+
+pub mod counter {
+    use super::*;
```
hashedone, line 15: I already roust you for that on the other review. Don't. Show explicitly what you import - this is not an additional layer of test abstraction; it is a standalone module implementing trait for the contract, which could as well be on a separate file. I don't see the structure of generated code this way.

sylvia/tests/querier.rs
```diff
@@ -0,0 +1,125 @@
...
+            self.count.save(ctx.deps.storage, &new_count)?;
+            Ok(Response::new())
+        }
+
+        #[msg(exec)]
+        fn copy_count(&self, ctx: ExecCtx, contract_addr: Addr) -> StdResult<Response> {
+            let other_count = counter::Remote::new(contract_addr)
+                .querier(&ctx.deps.querier)
+                .count()?
+                .count;
```
hashedone, line 53: I like what it is called. However, please, move to create `Remote::new(contract_addr)` to the instantiation and store it in the contract. This way, we show a crucial use case we want to support - keeping `Remote` as the contract state and later reaching it. Here you would just load `Remote` from the state.

hashedone (commented):
One thing is missing here. In generated MT utilities, we add functions on Queries. I would say, that it would be beneficial to use Queriers in them instead of reusing already generated utilities (to reduce maintenance cost).

I like what is getting generated and how is used, but I don't like the implementation structure.

sylvia-derive/src/utils.rs
```diff
@@ -1,14 +1,43 @@
...
 
+pub type ItemsIterator<'a> =
+    Box<dyn Iterator<Item = (&'a Vec<Attribute>, &'a Signature, Span)> + 'a>;
+
+/// Trait for extracting attributes and signature of the methods from `ItemImpl` and `ItemTrait`
+/// In most cases these two parameters are being used for preprocessing
+/// so to unify the logic we can use this trait
+pub trait Items<'a> {
+    fn items(&'a self) -> ItemsIterator;
+}
```
hashedone, line 21: I really like the implementation, I don't like naming. `Items` is a very generic name; it is not very clear what we are talking about. I know that here it is either `TraitItem` or `ImplItem`, but it is only because I know syn pretty well, but it doesn't really spell out what is its role in the codebase. Especially since the iterator inner type is also called... `Item`, which makes the name very confusing.

The `Item` in our case is something like `VariantDesc`, or maybe `MessageDesc` - something along those lines (you can come up with something with more sense, but something describing the role of it in the code).

Therefore the first thing I would do is to create the proper type for an "item". like `struct VariantDesc { ... }` containing fields for your tuple. This way, you never need to remember the field's order, and it is a generally better idea.

Additionally, as you keep `Span` in the type, it is probably a good idea to implement [syn::spanned::Spanned](https://docs.rs/syn/latest/syn/spanned/trait.Spanned.html) on the type - it looks natural.

The question here is if the items on the struct should be public. The `Span` for sure not - you should access it via the syn trait you implement on this.  However, you can either consider the `VariantDesc` type as just a data passing type and keep fields public, or you can figure out what kind of context it is used for and maybe provide proper functions on it. I would always prefer the second unless it ends up just implementing getters for its fields - but even in such a case keeping proper harmonization ends up being beneficial (as you almost always don't want to expose easy ways to modify fields after they are generated). Maybe I will give you a hint of how to work on it later and comment again.

Then I would rename the `Items` type to something like `IntoVariantDescs`, but implement it slightly differently. Here you used just `Box<dyn Iterator<...>>`, which is ok, but it is not very flexible - sometimes you want to provide the exact iterator type in implementation to make it better runtime efficient.

Therefore it should have an associated type named `Iter: Iterator<Item=VariantDesc>`, which is then used as the return value. You will still make it `Box<dyn Iterator<...>>` (or even use your alias, which is fine) in this context, but it gives you more flexibility and also prepares you for the upcoming [#63063](https://github.com/rust-lang/rust/issues/63063), when you will just change `Box<dyn Iterator<...>>` into `impl Iterator<...>` and you will get monopolization for free. In general you are creating the alternative [IntoIterator](https://doc.rust-lang.org/std/iter/trait.IntoIterator.html).

The tempting idea would be to not even create this trait, but instead, just implement the `IntoIterator` trait directly on `ItemImpl` and `ItemTrait`, but you cannot do it because of the orphan rule (and it would be also less verbose which is not ideal).

sylvia/tests/querier.rs
```diff
@@ -0,0 +1,131 @@
...
+            let other_count = remote.querier(&ctx.deps.querier).count()?.count;
+            self.count.save(ctx.deps.storage, &other_count)?;
+            Ok(Response::new())
+        }
+    }
+}
+
+pub struct CounterContract<'a> {
+    pub(crate) count: Item<'static, u64>,
+    pub(crate) remote: Item<'static, Remote<'a>>,
```
hashedone, line 63: This is unnecessary. Tests are not exposed from the library anyway, so if you want to make those fields public, just make them fully public. This `pub(crate)` immediately makes me look for a case when the `CounterContract` type leaks from the library which would be bad itself.

tarpaulin-report.html
hashedone, line 1: Good removing this file.

hashedone (commented):
Far better than the previous iteration, with some additional structure-wise comments.

sylvia-derive/src/message.rs
```diff
@@ -595,6 +587,112 @@ impl<'a> MsgVariant<'a> {
...
+        #[cfg(not(tarpaulin_include))]
+        {
+            quote! {
+                fn #variant_name(&self, #(#parameters),*) -> Result< #return_type, #sylvia:: cw_std::StdError>;
+            }
+        }
+    }
+}
+
+pub struct MsgVariants<'a>(Vec<MsgVariant<'a>>);
```
jawoznia, line 636: I'm not sure if this is a good approach to generate Querier directly from MsgVariants.

Later (in https://github.com/CosmWasm/sylvia/issues/28#issuecomment-1549648466) when we will implement `From` on the `BoundQuerier` we will need to pass the `interfaces` to `emit_querier` or source and parse it there inside. It doesn't fit in `new` so this would be a double pass of source inside the `MsgVariants`.

We could introduce a new struct, f.e. `ParsedInput` or however we would call it that would hold all the attributes, messages, error type and so on. This would reduce the preparsing which we currently do in multiple places: `multitest`, `EnumMessage`, `ContractEnumMessage`.
This way we could use this struct (in some separate PR) in other Structs generating code and extract all the commons to be generated by the `ParsedInput` to avoid duplications and reduce maintanence cost.

sylvia-derive/src/as_variant_descs.rs
hashedone, line 1: I would definitely name this file `variant_desc`, as the focus point is the variant description structure, the `as_...` trait is here to provide a creation method of this type.

sylvia-derive/src/message.rs
hashedone: The thing is - whatever is generated able from msg variants itself should be generated from here so we do not expose fields. What you are mentioning is a matter of the responsibility split.

When you need to emit `From` implementation on `BoundQuerier`, you can do it in another, maybe a new type, as it does not need access to message variants (it needs only the modules where "forms" are implemented).

Same thing when you implement traits on queries - maybe you will need some additional data, but only for the `impl` block header, not for functions - you will be able to emit functions implementation from `MsgVariants`, which would be passed (by ref) to some additional structure, and then the this additional structure will generate inner implementation calling some `Msgvariants` emit-family function, and then it would add some context (like `impl` block itself).

sylvia/tests/querier.rs
```diff
@@ -0,0 +1,131 @@
...
+    use cosmwasm_std::{Response, StdError, StdResult};
+    use sylvia::types::{ExecCtx, QueryCtx};
+    use sylvia::{contract, interface};
+
+    use crate::CountResponse;
+
+    #[interface]
+    pub trait Counter {
+        type Error: From<StdError>;
+
```
hashedone, line 23: Look, here you can keep your whitelines ;)

hashedone (approved):
LGTM.

Some minor comments, but please apply them.

PR #123, opened 2023-05-18
Disabling tarpaulin for quote

jawoznia:
Tarpaulin marks all `quote!` calls as not covered. This change aims to ignore them to stop false degrading coverage of the code


hashedone (approved):
LGTMhttps://www.youtube.com/watch?v=iYQ6aq85EFg

PR #127, opened 2023-05-24
Bound querier conversion

jawoznia:
PR3 of #28

### DIFF
--- a/sylvia-derive/src/message.rs
+++ b/sylvia-derive/src/message.rs
@@ -660,7 +660,38 @@ impl<'a> MsgVariants<'a> {
         Self(variants)
     }
 
-    pub fn emit_querier(&self) -> TokenStream {
+    pub fn emit_from_impls(&self, attrs: &'a [Attribute]) -> Vec<TokenStream> {
+        let sylvia = crate_module();
+
+        let from_impls: Vec<_> = attrs
+            .iter()
+            .filter(|attr| attr.path.is_ident("messages"))
+            .filter_map(|attr| {
+                let interface = match ContractMessageAttr::parse.parse2(attr.tokens.clone()) {
+                    Ok(interface) => interface,
+                    Err(err) => {
+                        emit_error!(attr.span(), err);
+                        return None;
+                    }
+                };
+
+                Some(interface)
+            })
+            .map(|interface| {
+                let ContractMessageAttr { module, .. } = interface;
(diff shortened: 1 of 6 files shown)


sylvia-derive/src/message.rs
```diff
@@ -682,6 +682,24 @@ impl<'a> MsgVariants<'a> {
...
+                        self.contract
+                    }
+
+                    pub fn new(contract: &'a #sylvia ::cw_std::Addr, querier: &'a #sylvia ::cw_std::QuerierWrapper<'a, C>) -> Self {
+                        Self {contract, querier}
+                    }
+
+                    pub fn borrowed(&self, querier: &'a #sylvia ::cw_std::QuerierWrapper<'a, C>) -> Self {
+                        Self::new(self.contract, querier)
+                    }
```
hashedone, line 700: Ok, this one is on me as I messed up signature, but we want this to be:

```rust
pub fn new(contract: &'a #sylvia ::cw_std::Addr, querier: &'a #sylvia ::cw_std::QuerierWrapper<'a, C>) -> Self {
    Self {contract, querier}
}
```

So basically, your `new` should be called borrowed.

sylvia-derive/src/message.rs
```diff
@@ -682,6 +682,24 @@ impl<'a> MsgVariants<'a> {
...
+                        self.querier
+                    }
+
+                    pub fn contract(&self) -> &'a #sylvia ::cw_std::Addr {
+                        self.contract
+                    }
+
+                    pub fn new(contract: &'a #sylvia ::cw_std::Addr, querier: &'a #sylvia ::cw_std::QuerierWrapper<'a, C>) -> Self {
+                        Self {contract, querier}
+                    }
```
hashedone, line 696: I don't think we ever need this function, but if so, then the signature is bad - `Addr` for this should be taken by owned value.

The reason this is probably needed is that if we store the address in the contract, then we probably store it as `Remote`, and we want to use the `.querier()` function to create a `BoundQuerier`. We use a `borrowed(...)` function only to make a shorthand to `Remote::borrowed(addr).querier(querier)`, but then I would prefer it to be called `borrowed` just like in `Remote`.

hashedone (approved):
LGTM, except their constructors are not exactly proper (consistency) - just fix them and then merge.

PR #129, opened 2023-05-24
entry_points no longer generate CONST Contract

jawoznia:
This removes requirement of creating `const fn new` method for all contracts


hashedone (approved):
LGTM

Issue #113, opened 2023-05-05
Generate Extension traits providing trait functions on contract queries


hashedone:
Part of #28

hashedone:
Closed by #141

Issue #112, opened 2023-05-05
Generate queriers for contracts


hashedone:
Part of #28

PR #141, opened 2023-05-26
Querier transparent calls

### LINKED ISSUE #28: Generate queriers for interfaces

jawoznia:
PR4
close  #28


jawoznia:
close #113 
close #112 
close #111

sylvia-derive/src/input.rs
```diff
@@ -243,4 +220,28 @@ impl<'a> ImplInput<'a> {
...
+        &self,
+        interfaces: &Interfaces,
+        variants: MsgVariants<'a>,
+    ) -> TokenStream {
+        let trait_module = interfaces
+            .interfaces()
+            .first()
+            .map(|interface| {
+                let module = &interface.module;
+                quote! { #module ::}
```
hashedone, line 234: I really don't like adding this `::` to the module - it is a minor thing, but on usage, i prefer `#module :: Item` far more than `#module Item`, as it is more obvious that module is some module path (not usage of it looks like it could be some `let` or another keyword). I know you are saving some common keystrokes, but I find it impacting readability too much.

sylvia-derive/src/message.rs
```diff
@@ -707,6 +707,29 @@ impl<'a> MsgVariants<'a> {
...
+
+        let methods_impl = variants
+            .iter()
+            .filter(|variant| variant.msg_type == MsgType::Query)
+            .map(|variant| variant.emit_querier_impl(trait_module));
+
+        #[cfg(not(tarpaulin_include))]
+        {
+            quote! {
+                impl <'a, C: #sylvia ::cw_std::CustomQuery> #trait_module Querier for #contract_module BoundQuerier<'a, C> {
```
hashedone, line 727: Like here, see difference between `#sylvia ::cw_std::CustomQuery` (which obviously is the module where `cw_std`) and `#trait_module Querier` where it looks like it is some strange querier-prefix - not very obvious what to expect there to me.

sylvia-derive/src/parser.rs
```diff
@@ -20,7 +20,7 @@ pub struct InterfaceArgs {
 /// Parser arguments for `contract` macro
 pub struct ContractArgs {
     /// Module name wrapping generated messages, by default no additional module is created
-    pub module: Option<Ident>,
+    pub module: Option<Path>,
```
hashedone, line 23: :+1:

sylvia/tests/dispatching.rs
```diff
@@ -1,38 +1,49 @@
...
-    fn no_args_query(&self, ctx: QueryCtx) -> Result<EmptyQueryResponse, Self::Error>;
-
-    #[msg(query)]
-    fn argumented_query(&self, ctx: QueryCtx, user: Addr) -> Result<QueryResponse, Self::Error>;
+mod interface {
+    use cosmwasm_std::{Addr, Decimal, Response, StdError};
+    use sylvia::{
+        interface,
+        types::{ExecCtx, QueryCtx},
+    };
```
hashedone, line 16: Split it per module for consistency:

```rust
use sylvia::interface;
use sylvia::types::{ExecCtx, QueryCtx};
```

hashedone (approved):
Approved with comments, please apply small ones, the more refactor-related: please acknowledge and leave some track of them. In general I like how it is done and the API we god.

sylvia-derive/src/input.rs
jawoznia: I also don't like it but here I had to do it because we need this module to be used also for QueryMsg later on.
We could extract the module into TokenStream and then merge with QueryMsg and Querier

contracts/cw20-base/src/multitest/receiver.rs
```diff
@@ -18,7 +18,7 @@ pub trait Receiver {
     ) -> Result<Response, Self::Error>;
 }
 
-#[contract]
+#[contract(module=crate::multitest::receiver_contract)]
```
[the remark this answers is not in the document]
jawoznia: me too

sylvia-derive/src/input.rs
jawoznia: changed so that Option<&Path> is being passed to these methods and inside we merge the modules with proper types.

sylvia-derive/src/input.rs
```diff
@@ -159,48 +140,44 @@ impl<'a> ImplInput<'a> {
...
         if is_trait {
-            return multitest_helpers;
+            let querier_bound_for_impl = self.emit_querier_for_bound_impl(&interfaces, variants);
+
+            #[cfg(not(tarpaulin_include))]
+            return quote! {
+                #multitest_helpers
+
+                #querier_bound_for_impl
+            };
```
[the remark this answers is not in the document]
jawoznia: Done under this PR

PR #142, opened 2023-05-26
Result emiting simplification

jawoznia:


sylvia-derive/src/parser.rs
```diff
@@ -123,21 +123,18 @@ impl MsgType {
...
-        match (self, msg_type) {
-            (Exec, Some(msg_type))
-            | (Instantiate, Some(msg_type))
-            | (Migrate, Some(msg_type))
-            | (Reply, Some(msg_type)) => {
+        match self {
+            Exec | Instantiate | Migrate | Reply => {
                 quote! {
-                    std::result::Result<#sylvia ::cw_std::Response<#msg_type>, #err_type>
+                    std::result::Result<#sylvia ::cw_std::Response #msg_type, #err_type>
```
hashedone, line 134: Again, don't like it. As I pointed out module has terminated `::`, so it's usage doesn't look like rust code immediately, here `#msg_type` is not just a type, it is not a valid entity - contains additional characters.

However I like what you did here in general, it can be simplified, by changing lines 126-129 to:

```rust
let response_type = msg_type
            .as_ref()
            .map(|ty| quote! { Response< #ty > })
            .unwrap_or_else(|| Response);
```

Maybe `Response` is "unnecessarily" repeated, but the entity is far more logical. You then just emit:

```rust
std::result::Result<#sylvia ::cw_std:: #response_type, #err_type>
```

hashedone (approved):
That is a very good idea, slight comment on execution to improve readability.

PR #143, opened 2023-05-26
Querier and entry_point macro described

jawoznia:


README.md
```diff
@@ -88,13 +88,13 @@ The next step is to create an instantiation message for the contract we have:
...
 use sylvia::contract;
-use sylvia::types::ExecCtx;
+use sylvia::types::InstantiateCtx;
 use sylvia::cw_std::{StdResult, Response};
 
 #[contract]
 impl MyContract {
     #[msg(instantiate)]
-    pub fn instantiate(&self, _ctx: ExecCtx) -> StdResult<Response> {
+    pub fn instantiate(&self, _ctx: InstantiateCtx) -> StdResult<Response> {
```
hashedone, line 97: BTW - that was working. That is why I am against type aliasing for those types. Also errors are not perfect. We need to reconsider that.

README.md
```diff
@@ -297,14 +311,19 @@ result type has to implement `Into<ContractError>`, where `ContractError` is a c
...
+Again entry point is already generated and ready to be used:
 
 ```rust
-use sylvia::cw_std::{Deps, Binary};
-
-#[entry_point]
-fn query(deps: Deps, env: Env, msg: ContractQueryMsg) -> Result<Binary, ContractError> {
-    msg.dispatch(&CONTRACT, (deps, env))
+pub mod entry_points {
+    #[entry_points]
```
hashedone, line 319: This is not valid.

I am not sure if it is a point to show entry points again and again, reading it I might be confused if I need to add this code or what.

README.md
```diff
@@ -410,6 +429,56 @@ for this interface (it would be used in generated code to provide proper enum va
...
+```
+
+`messages` is the attribute for the `contract` macro. We can use it both when implementing contract
+and when implementing an interface on a contract.
+In case of the implementation of a trait it is only needed if the trait is defined in different
+module. Otherwise it should be omitted.
+For the contract implementation it is mandatory for the functionality of an implemented trait 
+to be part of a contract logic.
+For the second usage in example there should be at most one `messages` attribute used.
+In case of the first usage there can be multiple `messages` attributes used.
```
hashedone, line 480: I don't like this `first`/`second`, it requires my eyes to jump up to ensure ordering. Just `interface`/`contract`.

README.md
```diff
@@ -459,6 +528,93 @@ with `ExecMsg/QueryMsg` - the former is generated only for contract, not for int
...
+
+Your contract might be implemented such it will be communicating with some other contract regularly.
+In such case you might want to store it as a field in your Contract:
+
+```rust
+pub struct MyContract<'a> {
+    counter: Item<'a, u64>,
+    members: Map<'a, &'a Addr, Empty>,
+    // Added
+    remote: Item<'a, Remote<'a>>,
```
hashedone, line 595: No, that is wrong. This is not a proper lifetime, `'a` is a lifetime for the name of an item; here we can store only the `'static` remote. If it works, it works by pure accident.

hashedone (approved):
Some comments - to be merged when fixed.

README.md
jawoznia: I agree. Because in case of compilation error it will point to ExecCtx instead of InstantiateCtx it should be changed to separate types

hashedone (approved):
LGTM

README.md
ethanfrey: I didn't realize there were different types, as ExecCtx worked here

ethanfrey: I didn't realize there were different types, as ExecCtx worked here

README.md
```diff
@@ -410,6 +402,66 @@ for this interface (it would be used in generated code to provide proper enum va
...
+...
+}
+```
+
+`module` is meant to be used when implementing interface on the contract. It's purpose
+is to inform `sylvia` where is the contract defined. If the contract is implemented in the same
+scope this attribute can and should be omitted.
+
+```rust
+#[entry_point]
```
ethanfrey, line 421: Nice to add a line about `entry_point` here as well (it is above, maybe just copy)

Issue #140, opened 2023-05-25
Fully support returning Response<CustomMsg> 


ethanfrey:
I got some code to compile with Sylvia 0.4.0 and no `mt` returning Response<CustomMsg> from instantiate (and in lib.rs)

When I updated to Sylvia 0.4.2 with `mt` flag enabled, it returned a compiler error.

The proc macros for the contract itself seem to handle it, seems to be more about mt.

ethanfrey:
My code

```rust
    #[msg(instantiate)]
    pub fn instantiate(
        &self,
        ctx: InstantiateCtx,
        denom: String,
    ) -> Result<Response<VirtualStakeCustomMsg>, ContractError> {
     ///
   }
```

`Cargo.toml`:

```toml
[dev-dependencies]
sylvia = { workspace = true, features = ["mt"]  }
```

The error message on `cargo test` (`cargo build` is fine):

```
error[E0308]: mismatched types
  --> contracts/consumer/virtual-staking/src/contract.rs:39:1
   |
39 | #[contract]
   | ^^^^^^^^^^^
   | |
   | expected `Result<Response, Error>`, found `Result<Response<...>, ...>`
   | expected `Result<Response, sylvia::anyhow::Error>` because of return type
   |
   = note: expected enum `Result<Response<cosmwasm_std::Empty>, sylvia::anyhow::Error>`
              found enum `Result<Response<VirtualStakeCustomMsg>, _>`
   = note: this error originates in the attribute macro `contract` (in Nightly builds, run with -Z macro-backtrace for more info)
```

hashedone:
Closes in favor of #157

PR #146, opened 2023-05-30
Response conversion

jawoznia:


sylvia/src/into_response.rs
```diff
@@ -0,0 +1,90 @@
...
+    fn into_msg(self) -> StdResult<SubMsg<C>>;
+}
+
+/// `SubMsg<Empty>` can be made into any `SubMsg<C>`
+impl<C> IntoMsg<C> for SubMsg<Empty> {
+    fn into_msg(self) -> StdResult<SubMsg<C>> {
+        let msg = match self.msg {
+            CosmosMsg::Wasm(wasm) => CosmosMsg::Wasm(wasm),
+            CosmosMsg::Bank(bank) => CosmosMsg::Bank(bank),
+            CosmosMsg::Staking(staking) => CosmosMsg::Staking(staking),
```
hashedone, line 14: It works by accident; you have `staking` feature flag enabled by something in cosmwasm-std. This line should be guarded with `#[cfg(feature = "staking")]` with this feature enabling staking on comwasm-std - see https://github.com/CosmWasm/cosmwasm/blob/main/packages/std/src/results/cosmos_msg.rs#L32-L33.

sylvia/src/into_response.rs
```diff
@@ -0,0 +1,90 @@
...
+            CosmosMsg::Distribution(distribution) => CosmosMsg::Distribution(distribution),
```
hashedone, line 15: As in prev comment - also enabled by the feature flag in `cosmwasm-std` (`staking` as well).

sylvia/src/into_response.rs
```diff
@@ -0,0 +1,90 @@
...
+                "Custom Empty message should not be sent",
+            ))?,
+            #[cfg(feature = "stargate")]
+            CosmosMsg::Ibc(ibc) => CosmosMsg::Ibc(ibc),
+            #[cfg(feature = "stargate")]
+            CosmosMsg::Stargate { type_url, value } => CosmosMsg::Stargate { type_url, value },
+            _ => Err(StdError::generic_err(format!(
+                "Unknown message variant: {:?}",
+                self
+            )))?,
```
hashedone, line 26: First, this should never happen; it means we don't convert some variants. I know it is needed because of `non_exhaustive` on `CosmosMsg`, but its firing means: "There was an update on cw-std, and you didn't align Sylvia to that yet" - and this should be explicit in the message as it is confusing now. Something like "Unknownown message variant: {:?}. Please make sure you are using the up-to-date Sylvia version, and if so please issue a bug on the Sylvia repository".

This error basically always mean - Sylvia is behind cw-std.

Other thing - never do `Err(err)?` - clippy warns about that (here clippy sometimes gets lost). Just `return StdError::generic_error(...)` (+ `.into()` if conversion is needed, eg. you use anyhow, but here you return `StdError` anyway).

sylvia/src/into_response.rs
```diff
@@ -0,0 +1,90 @@
...
+    fn into_response(self) -> StdResult<Response<T>>;
+}
+
+impl<T> IntoResponse<T> for Response<Empty> {
+    fn into_response(self) -> StdResult<Response<T>> {
+        let messages = self
+            .messages
+            .into_iter()
+            .map(|msg| msg.into_msg())
+            .collect::<StdResult<Vec<_>>>()?;
```
hashedone, line 48: I personally prefer:
```rust
let messages: Vec<_> = self
  // ...
  .collect::<Result<_, _>>()?;
```

To minimize turbofish. However, this is minor nitpicking.

hashedone (commented):
Pretty much ok, but some significant comments about how features are handled here.

Don't approve because I want to double-check after you fix them.

jawoznia:
part of #140

hashedone (approved):
LGTM

PR #149, opened 2023-06-01
Response conversion in generated code

jawoznia:
#140


jawoznia:
close in favor of #156

PR #150, opened 2023-06-02
Parametrized App, ExecProxy and MigrateProxy

jawoznia:
#140

### DIFF
--- a/sylvia/src/multitest.rs
+++ b/sylvia/src/multitest.rs
@@ -1,13 +1,51 @@
 use std::cell::{Ref, RefCell, RefMut};
+use std::fmt::Debug;
 use std::marker::PhantomData;
 
-use cosmwasm_std::{Addr, BlockInfo, Coin};
-use cw_multi_test::Executor;
+use cosmwasm_std::{
+    Addr, Api, BlockInfo, Coin, CustomQuery, Empty, GovMsg, IbcMsg, IbcQuery, Storage,
+};
+use cw_multi_test::{
+    AppBuilder, BankKeeper, DistributionKeeper, Executor, FailingModule, Router, StakeKeeper,
+    WasmKeeper,
+};
+use schemars::JsonSchema;
+use serde::de::DeserializeOwned;
 use serde::Serialize;
 
-#[derive(Default)]
-pub struct App {
-    pub app: RefCell<cw_multi_test::App>,
+pub struct App<ExecC = Empty, QueryC = Empty> {
+    pub app: RefCell<cw_multi_test::BasicApp<ExecC, QueryC>>,
+}
+
+impl Default for App {
+    fn default() -> Self {
+        Self::new(cw_multi_test::BasicApp::default())
+    }
+}
+
+/// Creates new default `App` implementation working with customized exec and query messages.
+/// Outside of `App` implementation to make type elision better.
+pub fn custom_app<ExecC, QueryC, F>(init_fn: F) -> App<ExecC, QueryC>
+where
+    ExecC: Debug + Clone + PartialEq + JsonSchema + DeserializeOwned + 'static,
+    QueryC: Debug + CustomQuery + DeserializeOwned + 'static,
+    F: FnOnce(
+        &mut Router<
+            BankKeeper,
+            FailingModule<ExecC, QueryC, Empty>,
+            WasmKeeper<ExecC, QueryC>,
+            StakeKeeper,
+            DistributionKeeper,
+            FailingModule<IbcMsg, IbcQuery, Empty>,
+            FailingModule<GovMsg, Empty, Empty>,
+        >,
+        &dyn Api,
+        &mut dyn Storage,
+    ),
+{
+    App {
+        app: RefCell::new(AppBuilder::new_custom().build(init_fn)),
+    }
 }
 
 impl App {
@@ -39,24 +77,28 @@ impl App {
 }
 
 #[must_use]
-pub struct ExecProxy<'a, 'app, Error, Msg>
+pub struct ExecProxy<'a, 'app, Error, Msg, ExecC = Empty, QueryC = Empty>
 where
     Msg: Serialize + std::fmt::Debug,
     Error: std::fmt::Debug + std::fmt::Display + Send + Sync + 'static,
+    ExecC: Debug + Clone + PartialEq + JsonSchema + DeserializeOwned + 'static,
+    QueryC: Debug + CustomQuery + DeserializeOwned + 'static,
 {
     funds: &'a [Coin],
     contract_addr: &'a Addr,
     msg: Msg,
-    app: &'app App,
+    app: &'app App<ExecC, QueryC>,
     phantom: PhantomData<Error>,
 }
 
-impl<'a, 'app, Error, Msg> ExecProxy<'a, 'app, Error, Msg>
+impl<'a, 'app, Error, Msg, ExecC, QueryC> ExecProxy<'a, 'app, Error, Msg, ExecC, QueryC>
 where
     Msg: Serialize + std::fmt::Debug,
     Error: std::fmt::Debug + std::fmt::Display + Send + Sync + 'static,
+    ExecC: Debug + Clone + PartialEq + JsonSchema + DeserializeOwned + 'static,
+    QueryC: Debug + CustomQuery + DeserializeOwned + 'static,
 {
-    pub fn new(contract_addr: &'a Addr, msg: Msg, app: &'app App) -> Self {
+    pub fn new(contract_addr: &'a Addr, msg: Msg, app: &'app App<ExecC, QueryC>) -> Self {
         Self {
             funds: &[],
             contract_addr,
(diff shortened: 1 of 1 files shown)


sylvia/src/multitest.rs
```diff
@@ -1,13 +1,51 @@
...
+
+impl Default for App {
+    fn default() -> Self {
+        Self::new(cw_multi_test::BasicApp::default())
+    }
+}
+
+/// Creates new default `App` implementation working with customized exec and query messages.
+/// Outside of `App` implementation to make type elision better.
+pub fn custom_app<ExecC, QueryC, F>(init_fn: F) -> App<ExecC, QueryC>
```
hashedone, line 28: I know it is how it is done in multitest, but did you try just create an `App::custom(...)` function for more consistent API?

sylvia/src/multitest.rs
```diff
@@ -39,24 +77,28 @@ impl App {
...
     phantom: PhantomData<Error>,
 }
 
-impl<'a, 'app, Error, Msg> ExecProxy<'a, 'app, Error, Msg>
+impl<'a, 'app, Error, Msg, ExecC, QueryC> ExecProxy<'a, 'app, Error, Msg, ExecC, QueryC>
 where
     Msg: Serialize + std::fmt::Debug,
     Error: std::fmt::Debug + std::fmt::Display + Send + Sync + 'static,
+    ExecC: Debug + Clone + PartialEq + JsonSchema + DeserializeOwned + 'static,
+    QueryC: Debug + CustomQuery + DeserializeOwned + 'static,
```
hashedone, line 99: ...and here a bound of `App: Executor<ExecC>`

hashedone (commented):
I know you wanted to simplify, but we cannot go with `BasicApp` - we have to go for full `App` supporting custom modules

sylvia/src/multitest.rs
```diff
@@ -1,13 +1,53 @@
...
+    MockStorage,
+    FailingModule<ExecC, QueryC, Empty>,
+    WasmKeeper<ExecC, QueryC>,
+    StakeKeeper,
+    DistributionKeeper,
+    FailingModule<IbcMsg, IbcQuery, Empty>,
+    FailingModule<GovMsg, Empty, Empty>,
+>;
+
+// This exist to silence `clippy::type-complexity`.
```
hashedone, line 29: This exists to improve readability; this comment is not needed. Clippy points it out to you because it affects code clarity.

sylvia/src/multitest.rs
```diff
@@ -1,13 +1,53 @@
...
+    Api = MockApi,
+    Storage = MockStorage,
+    Custom = FailingModule<Empty, Empty, Empty>,
+    Wasm = WasmKeeper<Empty, Empty>,
+    Staking = StakeKeeper,
+    Distr = DistributionKeeper,
+    Ibc = FailingModule<IbcMsg, IbcQuery, Empty>,
+    Gov = FailingModule<GovMsg, Empty, Empty>,
+> {
+    pub app: CWApp<Bank, Api, Storage, Custom, Wasm, Staking, Distr, Ibc, Gov>,
```
hashedone, line 44: I am pretty sure it is not what you want to keep it. You are getting into general hell.

This should most probably be:

```rust
pub struct App<MtApp> {
  app: RefCell<MtApp>,
}
```

So you don't care about the app's detailed type for most cases. Then what you do is you trait-bout the `MtApp` type depending on context.

And for the sake of whatever you feel fitting... stop creating public fields. Public fields should be only for data transfer types and nothing else. Particularly not for logic-wrapping types.

Rule of thumb - only create public fields if you just want type aggregating some common data which otherwise would be a tuple but you want your fields named. Otherwise always private.

sylvia/src/multitest.rs
```diff
@@ -1,13 +1,53 @@
...
+    FailingModule<ExecC, QueryC, Empty>,
+    WasmKeeper<ExecC, QueryC>,
+    StakeKeeper,
+    DistributionKeeper,
+    FailingModule<IbcMsg, IbcQuery, Empty>,
+    FailingModule<GovMsg, Empty, Empty>,
+>;
+
+// This exist to silence `clippy::type-complexity`.
+type CWApp<Bank, Api, Storage, Custom, Wasm, Staking, Distr, Ibc, Gov> =
```
hashedone, line 30: Most probably do not need this if you are properly trait-bound, as you should rarely construct the whole `App` type (except maybe creating `BasicApp<_>` type? Not sure if it is needed, but maybe.

sylvia/src/multitest.rs
```diff
@@ -39,22 +103,24 @@ impl App {
...
 where
-    Msg: Serialize + std::fmt::Debug,
-    Error: std::fmt::Debug + std::fmt::Display + Send + Sync + 'static,
+    Msg: Serialize + Debug,
+    Error: Debug + Display + Send + Sync + 'static,
 {
     funds: &'a [Coin],
     contract_addr: &'a Addr,
     msg: Msg,
     app: &'app App,
```
hashedone, line 114: This `app` type is invalid - it is the default app type, so it never supports custom queries/messages. It has to be generic.

sylvia/src/multitest.rs
```diff
@@ -39,22 +103,24 @@ impl App {
...
 
-impl<'a, 'app, Error, Msg> ExecProxy<'a, 'app, Error, Msg>
+impl<'a, 'app, Error, Msg, ExecC> ExecProxy<'a, 'app, Error, Msg, ExecC>
 where
-    Msg: Serialize + std::fmt::Debug,
-    Error: std::fmt::Debug + std::fmt::Display + Send + Sync + 'static,
+    Msg: Serialize + Debug,
+    Error: Debug + Display + Send + Sync + 'static,
+    ExecC: Debug + Clone + JsonSchema + PartialEq + 'static,
+    cw_multi_test::App: Executor<ExecC>,
```
hashedone, line 123: And this trait bound is always false for `ExecC` not being `Empty`, as `cw_multi_test::App` is a well-defined type monomorphisation. What you need here is to make your `app` a generic type (let's call it `MtApp` to not confuse with local `App`), and then here have trait bound:
```rust
MtApp: Executor<ExecC>,
```
This single line also tells me that you didn't write a single test calling those functions with custom `ExecC`.

contracts/cw1-subkeys/src/multitest/tests.rs
```diff
@@ -42,8 +42,11 @@ fn get_contract_version_works() {
         .call(owner)
         .unwrap();
 
-    let version: ContractVersion =
-        query_contract_info(&app.app().wrap(), contract.contract_addr.to_string()).unwrap();
+    let version: ContractVersion = query_contract_info(
+        &(*app).borrow_mut().wrap(),
```
hashedone, line 46: Nah, `Deref` doesn't work here as we would like to - we don't want to expose `RefCell`. It was my bad; you were right here. `.app()` was better (and we need to keep it), but because queries are common let's implement the `https://docs.rs/cosmwasm-std/latest/cosmwasm_std/struct.QuerierWrapper.html` directly on our `App`, so we can provide the direct `wrap()` function on `Querier` (and let's rename it to `querier` as the `wrap` name without the context is confusing to me.

contracts/cw1-subkeys/src/multitest/tests.rs
```diff
@@ -124,7 +127,7 @@ mod allowance {
             .call(owner)
             .unwrap();
 
-        let height = app.block_info().height;
+        let height = (*app).borrow().block_info().height;
```
hashedone, line 130: Better to implement `block_info()` directly like before or use `.app()`; this Deref looks bad. It is for all the instances, would comment once.

sylvia/src/multitest.rs
```diff
@@ -1,62 +1,100 @@
...
 
-    pub fn app(&self) -> Ref<'_, cw_multi_test::App> {
-        Ref::map(self.app.borrow(), |app| app)
+impl<MtApp> Deref for App<MtApp> {
+    type Target = RefCell<MtApp>;
+
+    fn deref(&self) -> &Self::Target {
+        &self.app
     }
+}
```
hashedone, line 33: I know I suggested it, but I was hoping we can do defer to `MtApp`, derefing to `RefCell<MtApp>` doesn't really make sense - better migrate functions to our `App`, giving the possibility to access `App` with `.app()` and `.app_mut()`.

sylvia/src/multitest.rs
```diff
@@ -1,62 +1,100 @@
...
+        ),
+    {
+        App {
+            app: RefCell::new(cw_multi_test::custom_app(init_fn)),
+        }
     }
+}
 
-    pub fn block_info(&self) -> BlockInfo {
-        self.app.borrow().block_info()
```
hashedone, line 29: I am pretty sure you can just move this `block_info` to the very same impl block as `custom`. You can try if it works, and if so leave it here, if it requires more work just create a new issue from this task for the future.

sylvia/src/multitest.rs
```diff
@@ -1,62 +1,100 @@
...
-    pub fn update_block<F: Fn(&mut BlockInfo)>(&self, action: F) {
-        self.app.borrow_mut().update_block(action)
+    pub fn app_mut(&self) -> RefMut<'_, MtApp> {
+        RefMut::map(self.app.borrow_mut(), |app| app)
     }
 }
 
 #[must_use]
-pub struct ExecProxy<'a, 'app, Error, Msg>
+pub struct ExecProxy<'a, 'app, Error, Msg, MtApp, ExecC = Empty>
```
hashedone, line 78: I am struggling if it is a good idea to have this `Empty` default for `ExecC` - does it save much? If we have it - why not make `MtApp` default to `mt::BasicApp<Empty>`? I kind of see that `MtApp` have this additional `QueryC` you will sometimes change... Therefore I am not super sure. I think I would just get rid of default here as it saves you a bit right now, but in the future, you will anyway always spell this whole type in generated code, and the type is not to be spelled out by the user anyway. Removing the default here would save you some trouble working with errors of "`MtApp doesn't implement the `Executor<ExecC>`, but it implements `Executor<SomeOther`", bla bla bla, which would happen if you would have non-default `MtApp` not matching `Empty` `ExecC`, and you forget to specify `ExecC`. If you undefault it, you will have an error of missing generic and things should be a bit easier to maintain.

hashedone (approved):
Ok, deref doesn't work; thank you for trying it out. Good job on proving me wrong.

The generic `App` looks far better now; it probably will be good to go. Minor comment about default.

Please get rid of the deref impl, and fall back to `.app()`, `app_mut()`, other comments are optional. If you decide one is to be done but is a bit to much work, please make an issue out of the comment for the future.

I give you approve as those are changes I believe you will solve properly.

contracts/cw1-subkeys/src/multitest/tests.rs
jawoznia: Requires https://github.com/CosmWasm/cw-multi-test/pull/43

contracts/cw1-subkeys/src/multitest/tests.rs
jawoznia: tracked under #154

PR #172, opened 2023-06-09
fix: Generate code using sylvia::cw_std instead of cosmwasm_std

### LINKED ISSUE #105: Fix usages of cosmwasm_std
Sylvia reexports cosmwasm_std, so it should use it instead of assuming user adds its as dependency.

jawoznia:
close #105


hashedone (approved):
LGTM

PR #176, opened 2023-06-14
feat: support for custom attribute in messages

jawoznia:


jawoznia:
close #159

sylvia-derive/src/message.rs
```diff
@@ -921,9 +922,19 @@ impl<'a> GlueMessage<'a> {
...
+                is_custom_msg,
+                ..
+            } = interface;
+
+            if *is_custom_msg && msg_ty == &MsgType::Exec
+            {
+                quote! { #contract_name :: #variant(msg) => Ok( #sylvia ::into_response::IntoResponse::into_response(msg.dispatch(contract, Into::into(ctx))?)?) }
+            } else {
+                quote! { #contract_name :: #variant(msg) => msg.dispatch(contract, Into::into(ctx)) }
+            }
```
hashedone, line 936: I don't like this decision. You introduce branching in every `interface`, while clearly, all interfaces are going to call the same branch.

It is not even a performance issue (technically might affect, but not really), but a readability issue - I need to think about it to understand that it is always the same branch taken.

Both of those ifs "bodies" shall be extracted to some function (even local closures), and then one should be selected before the iteration. The best is to maintain this function signature as `fn(&ContractMessageAttr) -> TokenStream` (not sure about result type - whatever is returned by `quote!`), then you simplify this to something like:

```rust
let emit_dispatch_arm = match *is_custom_msg && msg_ty == &MsgType::Exec {
  true => emit_dispatch_arm_resp_conversion,
  false => emit_default_dispatch_arm,
};
let dispatch_arms = interfaces.iter().map(emit_dispatch_arm);
```

Possibly even inlining the dispatch arms there.

sylvia-derive/src/parser.rs
```diff
@@ -181,6 +181,7 @@ pub struct ContractMessageAttr {
     pub exec_generic_params: Vec<Path>,
     pub query_generic_params: Vec<Path>,
     pub variant: Ident,
+    pub is_custom_msg: bool,
```
hashedone, line 184: It is not `is`. It may be `has`, but in general, bool fields in structures always looks suspicious to me, as they could be just calculated inline or like some content is missing. Here I am wondering if keeping the `path` of the `content` or at least its span may be useful for error reporting, but maybe not - unless we use it we may leave it this way. Bot not with this name.

hashedone (approved):
Seems decent, with comments.

sylvia-derive/src/message.rs
jawoznia: As discussed offline this is this is dispatch per interface and it's correct behavior. I can agree on changing it to match

PR #177, opened 2023-06-14
Support associated `ExecC` type on interfaces

### LINKED ISSUE #160: Implement detection of `ExecC` associate type on the traits
`type ExecC` on associated trait should cause the dispatcher of this `Exec` trait return `Result<Response<<Contract as InterfaceTrait>::ExecC>`.

jawoznia:
close #160


jawoznia:
close #161 
This change prioritize custom attribute over associated type

jawoznia:
Depends on #176

sylvia-derive/src/message.rs
```diff
@@ -203,6 +203,25 @@ impl<'a> EnumMessage<'a> {
...
 
+        let associated_exec = source.items.iter().find_map(|item| match item {
+            TraitItem::Type(ty) => {
+                if ty.ident == "ExecC" {
+                    Some(quote! { <C as #trait_name>::ExecC })
+                } else {
+                    None
+                }
+            }
+            _ => None,
```
hashedone, line 214: You can guard here:
```suggestion
            TraitItem::Type(ty) if ty.ident == "ExecC" =>
                Some(quote! { <C as #trait_name>::ExecC })
            _ => None,
```

sylvia-derive/src/message.rs
```diff
@@ -203,6 +203,25 @@ impl<'a> EnumMessage<'a> {
...
+        });
+
+        let resp_type = match associated_exec {
+            Some(exec) if !custom.is_msg() => exec,
```
hashedone, line 218: And here you remind yourself about match-guards :D

sylvia-derive/src/parser.rs
```diff
@@ -82,7 +82,7 @@ impl MsgType {
     }
 
     /// Emits type which should be returned by dispatch function for this kind of message
-    pub fn emit_result_type(self, msg_type: &Path, err_type: &Type) -> TokenStream {
+    pub fn emit_result_type(self, msg_type: impl ToTokens, err_type: &Type) -> TokenStream {
```
hashedone, line 85: Not sure about that - now you loos your type check that the entity passed here is something reasonable. I can pass the expression here. Logically - only `Path` and maybe a `Type` or an `Ident` can go here, but you can always make both `Ident` and `Type` into `Path` with `Into` traits.

sylvia-derive/src/parser.rs
```diff
@@ -315,19 +315,21 @@ pub fn parse_struct_message(source: &ItemImpl, ty: MsgType) -> Option<(&ImplItem
...
     fn default() -> Self {
         let sylvia = crate_module();
 
         Self {
-            msg: parse_quote!(#sylvia ::cw_std::Empty),
-            _query: parse_quote!(#sylvia ::cw_std::Empty),
+            msg: None,
+            _query: None,
             input_attr: None,
+            default_type: parse_quote! {#sylvia ::cw_std::Empty},
```
hashedone, line 332: This is not supposed to be a field - it is basically a constant. I know you cannot make it constant, but it is not something that can have other values, better to make it into `fn default_type() -> Path` and build on-demand.

sylvia-derive/src/parser.rs
```diff
@@ -370,7 +372,11 @@ impl<'a> Custom<'a> {
     }
 
     pub fn msg(&self) -> &Path {
-        &self.msg
+        self.msg.as_ref().unwrap_or(&self.default_type)
```
hashedone, line 375: Ok, I get it - you wanted to return a reference here, so you needed to keep it somewhere. Makes a little sense, but also seems a bit strange. This seems like an exact case for `https://docs.rs/lazy_static/latest/lazy_static/`.

sylvia-derive/src/parser.rs
```diff
@@ -370,7 +372,11 @@ impl<'a> Custom<'a> {
...
+    }
+
+    pub fn is_msg(&self) -> bool {
```
hashedone, line 378: It is not `is_msg`; it is `has_msg` if anything.

hashedone (approved):
Some comments here

sylvia-derive/src/parser.rs
jawoznia: `lazy_static` requires type to implement `Sync` which `Path` does not. I think none of the types in `syn` does.

hashedone: Yea, ok, I see this. I don't like this in general, but for now we can leave it here.

PR #179, opened 2023-06-16
Custom msg support

### LINKED ISSUE #157: Custom chain messages on sylvia contracts
We want to be able to able to use smart contacts on external chains, which requires three items:
1) Ability to return `CosmosMsg::Custom(...)` variant of `CosmosMsg` in the `Response` to send a cosmos-specific message by returning `Response<ExecC>` from the exec message.
2) Ability to query the chain with custom query by taking the `Deps<QueryC>` and `DepsMut<QueryC>` types as in context (which requires context to be generic over those).
3) Ability to handle the `sudo` entry point.

This issue covers only the 1) point of this.

## Returning the `Response<ExecC>` from the function

The clue of the issue is to allow returning a parametrized response from the messages on the contract. The example case:

```rust
#[contract]
impl MyContract {
  #[msg(instantiate)]
  fn instantiate(&self, ctx: InstantiateCtx, param: u64) -> Result<Response<OsmosisMsg>, ContractError> {
    let msg = OsmosisMsg::SomeVariant { param };
    let msg = CosmsoMsg::Custom(msg);
    let resp = Reponse::new()
      .add_message(msg);
    Ok(resp)
  }
}
```

It could be achieved by parsing the return type of every function returning `Response`, and if any of them returns parametrized variant, make an entry point returning the parametrized variant and make it the `ExecC` of the contract.

As tempting, this approach will be difficult to implement and vulnerable to edge cases. Example: what if one would like to create an alias of `type OsmoResp = Response<OsmosisMsg>`? Extracting the `ExecC` here from the macro position would be challenging.

Another problem - what if many functions are returning the `ExecC` from functions, but they differ? For now, we don't want to support any conversion between them, so this is an error, but it isn't easy to decide how to describe that clearly.

The solution here is to annotate the contract with what we want to be the contract `ExecC` on the contract implementation:

```rust
#[contract]
#[sv::custom(msg=OsmosisMsg)]
impl MyContract {
  #[msg(instantiate)]
  fn instantiate(&self, ctx: InstantiateCtx, param: u64) -> Result<Response<OsmosisMsg>, ContractError> {
    let msg = OsmosisMsg::SomeVariant { param };
    let msg = CosmsoMsg::Custom(msg);
    let resp = Reponse::new()
      .add_message(msg);
    Ok(resp)
  }
}
```

Now the entry points would be generated returning `Result<Response<OsmosisMsg>>`, and the error would be on invalid type returned from the dispatcher (we want to make sure the span for that points to the result type of the function).

When the contract has defined the `msg=...`, we require the user to always return a proper `Reponse<...>` from the contract. The reason for that is it is impossible to define the conversion which would work both as identity conversion and generic conversion from `Response<Empty>` to `Response<T>` due to conflicting implementation for `Response<Empty>` to `Response<Empty>`.

## Handling responses returned in interfaces

For now, we do not support any `ExecC` being returned for variants defined in interfaces, and we want to keep supporting this for compatibility. Messages generated for such traits would return the `Response<Empty>` type. However, we want to be able to implement them on our contracts. Let's consider such a code:

```rust
mod iface {
#[interface]
  pub trait IFace {
    type Error;
    
    #[exec]
    fn variant(&self, ctx: ExecCtx) -> Result<Response, Self::Error>;
  }
}

mod iface_impl {
  #[contract]
  #[messages(iface as IFace)]
  impl iface::IFace for MyContract {
    type Error = ContractError;
  
    #[exec]
    fn variant(&self, ctx: ExecCtx) -> Result<Response, ContractError> {
      Ok(Response::new())
    }
  }
}


#[contract]
#[sv::custom(msg=OsmosisMsg)]
#[messages(iface as IFace: custom(msg))]
impl MyContract {
  #[msg(instantiate)]
  fn instantiate(&self, ctx: InstantiateCtx, param: u64) -> Result<Response<OsmosisMsg>, ContractError> {
    let msg = OsmosisMsg::SomeVariant { param };
    let msg = CosmsoMsg::Custom(msg);
    let resp = Reponse::new()
      .add_message(msg);
    Ok(resp)
  }
}
```

We want this to work, and to make it work properly, we need to call the conversion function of `Response` of whatever got returned by interface msg `dispatch()` in the contract-level `dispatch()`. However the problem occurs - we can call the function only if there is a conversion to be performed, as our conversion function cannot support identity (for reasons already explained).

To solve this, we will call the conversion utility if and only if `custom(msg)` "option" is added to the corresponding `#[messages(...)]` argument.

## `ExecC` in interfaces

Interfaces should be chain-agnostic in the majority of cases. However, it should be possible to implement the message from the interface that chooses to send a custom message to the chain it is implemented on. The way we will approach this is to leverage the associated type with the `ExecC`:

```rust
mod iface {
#[interface]
  pub trait IFace {
    type Error;
    type ExecC;
    
    #[exec]
    fn variant(&self, ctx: ExecCtx) -> Result<Response<Self::ExecC>, Self::Error>;
  }
}

mod iface_impl {
  #[contract]
  #[messages(iface as IFace)]
  impl iface::IFace for MyContract {
    type ExecC = OsmosisMsg;
    type Error = ContractError;
  
    #[exec]
    fn variant(&self, ctx: ExecCtx) -> Result<Response<OsmosisMsg, ContractError> {
        let msg = OsmosisMsg::SomeVariant { param };
        let msg = CosmsoMsg::Custom(msg);
        let resp = Reponse::new()
          .add_message(msg);
        Ok(resp)
     }
  }
}

#[contract]
#[sv::custom(msg=OsmosisMsg)]
#[messages(iface as IFace)]
impl MyContract {
  #[msg(instantiate)]
  fn instantiate(&self, ctx: InstantiateCtx, param: u64) -> Result<Response<OsmosisMsg>, ContractError> {
    let msg = OsmosisMsg::SomeVariant { param };
    let msg = CosmsoMsg::Custom(msg);
    let resp = Reponse::new()
      .add_message(msg);
    Ok(resp)
  }
}
```

This way the trait can implementation can use any chain-specific custom messages. To achieve that we will look for the `ExecC` associated type definition, and if it exists, we will return `Result<Response<Self::ExecC>>` from the generated message dispatch. If there is no such a message, we will fall back to return just a `Result<Response, ContractError>`.

Nothing should change here regarding implementing a trait on contract - the user will need to specify the `ExecC`, but the Rust itself would enforce it. 

Note the lack of `#[custom(...)]` on the interface impl block - our trait already returns proper `Response` and we don't want to convert it as we do not support identity conversion!

## Chain-specific interfaces

Finally, I can imagine interfaces that are considered to work only on particular chains. As it doesn't seem to be the true case (I can always try to implement the same behavior on different chains), it might be a case this simplifies some contracts not bothering about the associated `ExecC` type. To support that, we want to allow the user to specify the `ExecC` on the trait directly with the `sv::custom(...)`:

```rust
mod iface {
  #[interface]
  #[sv::custom(msg=OsmosisMsg)]
  pub trait IFace {
    type Error;
    
    #[exec]
    fn variant(&self, ctx: ExecCtx) -> Result<Response<OsmosisMsg>, Self::Error>;
  }
}

#[contract]
#[messages(iface as IFace)]
impl iface::IFace for MyContract {
  type Error = ContractError;
  
  #[exec]
  fn variant(&self, ctx: ExecCtx) -> Result<Response<OsmosisMsg, ContractError> {
    Ok(Response::new())
  }
}

mod iface_impl {
  #[contract]
  #[sv::custom(msg=OsmosisMsg)]
  #[messages(iface as IFace)]
  impl MyContract {
    #[msg(instantiate)]
    fn instantiate(&self, ctx: InstantiateCtx, param: u64) -> Result<Response<OsmosisMsg>, ContractError> {
      let msg = OsmosisMsg::SomeVariant { param };
      let msg = CosmsoMsg::Custom(msg);
      let resp = Reponse::new()
        .add_message(msg);
      Ok(resp)
    }
  }
}
```

This way user can control exactly what type can be used as the `ExecC` in the response (it can even be the trait generic or something). This way we can make it a bit easier for a "project private" interface designed for a particular chain.

## Warning about non-chain-agnostic interfaces

Finally, most interfaces should be chain agnostic and defined with `ExecC`. Unfortunately - adding and maintaining this is another thing to remember, maintain, and document. We want Sylvia to help us with that, so whenever we see the interface with no `ExecC` or `#[custom(msg=...)]`, we assume it is an oversight or leftover after migration from previous Sylvia releases.  We want to support it (mostly to make migrations as flawless as possible), but we will warn about it something like:

```
The trait is missing the custom message definition, and returning any chain-specific custom messages on its implementation will not be possible.
Note: Consider adding the `type ExecC;` associated message in the trait definition, and then use it in the execution responses return type: `Result<Response<Self::ExecC>, _>`
Note: If it is intended and you don't want to allow executing custom messages from this trait, add the `#[sv::custom(cosmwasm_std::Empty)]` attribute on the trait.
```

## Multitest

All of the above has to have additional support in the multitest utilities. In particular - we need the contract helper to implement `Contract<ExecC>` instead of just `Contract`. Everything else should be completely transparent from MT PoV.

## Tasks split
- [x] #158 
- [x] #159 
- [x] #160
- [x] #161 
- [ ] #162

jawoznia:
close #157

### DIFF
--- a/sylvia/tests/custom_msg.rs
+++ b/sylvia/tests/custom_msg.rs
@@ -21,14 +21,13 @@ pub struct SomeResponse;
 
 mod some_interface {
     use cosmwasm_std::{Response, StdError, StdResult};
-    use sylvia::{
-        interface,
-        types::{ExecCtx, QueryCtx},
-    };
+    use sylvia::interface;
+    use sylvia::types::{ExecCtx, QueryCtx};
 
-    use crate::SomeResponse;
+    use crate::{MyMsg, SomeResponse};
 
     #[interface]
+    #[sv::custom(msg=MyMsg, query=MyQuery)]
     pub trait SomeInterface {
         type Error: From<StdError>;
 
@@ -38,7 +37,7 @@ mod some_interface {
 
         #[cfg(not(tarpaulin_include))]
         #[msg(exec)]
-        fn interface_exec(&self, ctx: ExecCtx) -> StdResult<Response>;
+        fn interface_exec(&self, ctx: ExecCtx) -> StdResult<Response<MyMsg>>;
     }
 
     impl SomeInterface for crate::MyContract {
@@ -48,14 +47,13 @@ mod some_interface {
             Ok(SomeResponse)
         }
 
-        fn interface_exec(&self, _ctx: ExecCtx) -> StdResult<Response> {
+        fn interface_exec(&self, _ctx: ExecCtx) -> StdResult<Response<MyMsg>> {
             Ok(Response::default())
         }
     }
 }
 
 #[contract]
-// Uncomment once into_response is called in dispatch
 // #[messages(some_interface as SomeInterface)]
 #[sv::custom(msg=MyMsg, query=MyQuery)]
 impl MyContract {
--- a/sylvia/tests/custom_msg.rs
+++ b/sylvia/tests/custom_msg.rs
@@ -2,7 +2,7 @@ use cosmwasm_std::{CustomMsg, CustomQuery, Response, StdResult};
 use schemars::JsonSchema;
 use serde::{Deserialize, Serialize};
 use sylvia::contract;
-use sylvia::types::{ExecCtx, InstantiateCtx};
+use sylvia::types::{ExecCtx, InstantiateCtx, MigrateCtx, QueryCtx};
 
 #[derive(Clone, PartialEq, Serialize, Deserialize, Debug, JsonSchema)]
 pub struct MyMsg;
@@ -10,13 +10,51 @@ pub struct MyMsg;
 impl CustomMsg for MyMsg {}
 
 #[derive(Clone, PartialEq, Serialize, Deserialize, Debug, JsonSchema)]
-struct MyQuery;
+pub struct MyQuery;
 
 impl CustomQuery for MyQuery {}
 
 pub struct MyContract;
 
+#[derive(Clone, PartialEq, Serialize, Deserialize, Debug, JsonSchema)]
+pub struct SomeResponse;
+
+mod some_interface {
+    use cosmwasm_std::{Response, StdError, StdResult};
+    use sylvia::{
+        interface,
+        types::{ExecCtx, QueryCtx},
+    };
+
+    use crate::SomeResponse;
+
+    #[interface]
+    pub trait SomeInterface {
+        type Error: From<StdError>;
+
+        #[msg(query)]
+        fn interface_query(&self, ctx: QueryCtx) -> StdResult<SomeResponse>;
+
+        #[msg(exec)]
+        fn interface_exec(&self, ctx: ExecCtx) -> StdResult<Response>;
+    }
+
+    impl SomeInterface for crate::MyContract {
+        type Error = StdError;
+
+        fn interface_query(&self, _ctx: QueryCtx) -> StdResult<SomeResponse> {
+            Ok(SomeResponse)
+        }
+
+        fn interface_exec(&self, _ctx: ExecCtx) -> StdResult<Response> {
+            Ok(Response::default())
+        }
+    }
+}
+
 #[contract]
+// Uncomment once into_response is called in dispatch
+// #[messages(some_interface as SomeInterface)]
 #[sv::custom(msg=MyMsg, query=MyQuery)]
 impl MyContract {
     #[allow(clippy::new_without_default)]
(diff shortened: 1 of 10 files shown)


sylvia/tests/custom_msg.rs
```diff
@@ -25,17 +162,78 @@ impl MyContract {
...
+    }
+
+    #[cfg(not(tarpaulin_include))]
+    #[msg(migrate)]
+    pub fn some_migrate(&self, _ctx: MigrateCtx) -> StdResult<Response<MyMsg>> {
         Ok(Response::default())
     }
 }
 
 #[cfg(test)]
```
hashedone, line 186:
```suggestion
#[cfg(all(test, feature = "mt"))]
```

hashedone:
Addresses https://github.com/confio/public-goods/issues/85

hashedone (approved):
LGTM

PR #188, opened 2023-06-30
Fix tarpaulin on ci

jawoznia:


.gitignore
```diff
@@ -2,3 +2,4 @@ target/**
 */target/**
 **/schema/**
 tarpaulin-report.html
+*.profraw
```
hashedone, line 5: Is this generated by something in project, or your private file? If your private thingy, then it shall go to your local ignored list

hashedone (approved):
LGTM, not sure about additional ignore.

.gitignore
jawoznia: this is generated when running tarpaulin with llvm engine

Issue #189, opened 2023-06-30
Support defining custom entry points


jawoznia:
# Motivation

As described in #139 although user can define custom `entry_points` and use them in the generated contract it is not possible to use it in generated `multitests`.

# Solution

We could either use macro attribute pointing to the custom entry_point function or add method attribute inside of the macro.
Macro attribute approach will be more flexible for the user.

## Add missing Context type

```rust
pub struct SudoCtx<'a, C: CustomQuery = Empty> {
    pub deps: DepsMut<'a, C>,
    pub env: Env,
}
```

## Detect attribute `sv::custom_entry_point` on contract

This attribute should be used both by `contract` macro to use the custom entry point in `multitest helpers` and by `entry_points` macro so that there will be no duplication after generation.

```rust
pub mod entry_points {
    #[entry_point]
    pub fn sudo(_deps: DepsMut, _env: Env) -> ... {
    }
}

#[cfg_attr(not(feature = "library"), entry_points)]
#[contract]
#[sv::custom_entry_point(sudo=entry_points::sudo)]
impl ReceiverContract {
    pub const fn new() -> Self {
        Self {}
    }

    #[msg(instantiate)]
    pub fn instantiate(&self, _ctx: InstantiateCtx) -> StdResult<Response> {
        todo!()    
    }
}
``` 

## Allow overriding all entry points

This is the last step which aim is to make `sylvia` more flexible for the user. In case of some missing feature or individual need user should be able to override all of the entry points.

# Alternative

Although support for `sudo` is not yet implemented we could allow defining `sudo` msg as it is done in case of `reply`. If detected `sylvia` will generate appropriate `entry_point` and implement `cw_multi_test::Contract` with this `entry_point`.

hashedone:
To make it complete - it should be possible to overwrite all of the entry points to possibly overwrite some sylvia behaviour in case of need (in particular: missing features).

jawoznia:
In case of overwriting entry points other than `sudo` it shouldn't cause any issues.

In case of `sudo` we still don't generate the `SudoMsg`. This causes an issue with dispatch.

For `cw_multitest::Contract` implementation on user contract we need to define the type of the incoming message.

For f.e. query we will cast the received byte array to generated `ContractQueryMsg' type.

```rust
fn query(
    &self,
    deps: sylvia ::cw_std::Deps<sylvia ::cw_std::Empty>,
    env: sylvia ::cw_std::Env,
    msg: Vec<u8>,
[...]

PR #192, opened 2023-07-07
feat: New type SudoCtx

jawoznia:
Part of #189


hashedone (approved):
It is not a full feat so I would prefer it to be done with the usage, but besides that LGTM

PR #193, opened 2023-07-07
feat: Override entry_points in multitest helpers

jawoznia:
Part of #189

### DIFF
--- a/sylvia-derive/src/multitest.rs
+++ b/sylvia-derive/src/multitest.rs
@@ -9,7 +9,10 @@ use syn::{FnArg, GenericParam, ImplItem, ItemImpl, ItemTrait, Pat, PatType, Path
 use crate::check_generics::CheckGenerics;
 use crate::crate_module;
 use crate::message::MsgField;
-use crate::parser::{parse_struct_message, ContractMessageAttr, Custom, MsgAttr, MsgType};
+use crate::parser::{
+    parse_struct_message, ContractMessageAttr, Custom, MsgAttr, MsgType, OverrideEntryPoint,
+    OverrideEntryPoints,
+};
 use crate::utils::{extract_return_type, process_fields};
 
 struct MessageSignature<'a> {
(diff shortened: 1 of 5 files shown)


sylvia-derive/src/multitest.rs
```diff
@@ -696,32 +702,64 @@ impl<'a> MultitestHelpers<'a> {
...
-        } else {
-            quote! {
-                #sylvia ::anyhow::bail!("reply not implemented for contract")
-            }
+        let reply_body = match override_entry_points.get_entry_point(MsgType::Reply) {
+            Some(entry_point) => entry_point.emit_multitest_dispatch(),
+            None => self
+                .reply
+                .as_ref()
+                .map(|reply| {
```
hashedone, line 753: I don't think that you need those external curly braces, so if it is not becasue of some formatting issue I would get rid of them.

hashedone (commented):
In general looks fine, but I am not sure what happens if you overwrite the entry point which has messages defined on the contract?

What should happen is that it should use the overwritten entry point, end every definition of message related to the entry point should emit a warning that entry point is overwritten.

What I would like at least is the test case for this situation - eg contract with execution overwritten and some `#[msg(exec)]` things, and then some MT verifying the overwritten one is executed.

sylvia-derive/src/multitest.rs
jawoznia: It is because of the formatting unfortunately

jawoznia:
> In general looks fine, but I am not sure what happens if you overwrite the entry point which has messages defined on the contract?
> 
> What should happen is that it should use the overwritten entry point, end every definition of message related to the entry point should emit a warning that entry point is overwritten.
> 
> What I would like at least is the test case for this situation - eg contract with execution overwritten and some `#[msg(exec)]` things, and then some MT verifying the overwritten one is executed.

Done, but I took another approach and let the user to decide. User can still use generated messages but need to provide own dispatch logic.

hashedone (approved):
Ok, at least I see it tested now. We will think about any warning later if we would see it becomes an issue.

Ready to merge

PR #198, opened 2023-07-20
chore: Remove version from workspace sylvia-derive dependency

jawoznia:
# Motivation

`Release-plz` didn't update workspace-dependencies. Tried also to add workspace-dependency for `cosmwasm-std` but with no success.

# Workaround

We can remove `sylvia-derive` version from deps and rely solely on path. In future we should come up with solution to automatize workspace dependencies update and migrate shared dependencies to workspace `Cargo.toml`


jawoznia:
> Why is that? DIdn't we want to move version entirely to per-project cargo?

You mean workspace cargo?
I made this change because version in `workspace.dependency` is not updated. For now it might be a WA.

jawoznia:
After offline talk agree this would damage crates.io release

PR #199, opened 2023-07-20
docs: Remove unnecessary slashes from CHANGELOG.md

jawoznia:
These slashes were added by auto formatter. They might cause an issue with git-cliff and cause it to add release notes in the middle of the file rather then on top.


hashedone (approved):
LGTM

Issue #201, opened 2023-07-26
Implement the #[sv::custom(query=CustomQuery)] attribute on the contract level


jawoznia:
Part of #16 
- [x] `#[sv::custom(query=CustomQuery)]` causes generated entry points and messages expect `Deps<CustomQuery>`
- [x] Contract MT helper implements `Contract<.., CustomQuery>`
- [x] All MT helpers generated work properly with generic `App` (with specified `QueryC`)
- [x] `#[messages(...: custom(query))]` on the contract implementation causes the conversion of deps dispatched to the interface

PR #202, opened 2023-07-26
feat: Support custom queries on contracts

jawoznia:
First and second part of #201


hashedone (approved):
LGTM.

What would be great is if we could provide some tests, where the contract actually queries the `CustomQuery` to prove it works - it should be possible by providing the proper module to the `App`. It's ok to add it in separate PR.

PR #210, opened 2023-08-01
feat: Support QueryC associated type on interface

### LINKED ISSUE #203: Implement detection of QueryC associated type on the traits
Part of #16 
type `QueryC` on associated trait should cause the dispatcher of this Query trait to accept `Deps<<Contract as InterfaceTrait>::QueryC>`

jawoznia:
closes #203

### DIFF
--- a/sylvia-derive/src/message.rs
+++ b/sylvia-derive/src/message.rs
@@ -1,8 +1,8 @@
 use crate::check_generics::CheckGenerics;
 use crate::crate_module;
 use crate::parser::{
-    parse_struct_message, ContractErrorAttr, ContractMessageAttr, Custom, MsgAttr, MsgType,
-    OverrideEntryPoint, OverrideEntryPoints,
+    parse_associated_custom_type, parse_struct_message, ContractErrorAttr, ContractMessageAttr,
+    Custom, MsgAttr, MsgType, OverrideEntryPoint, OverrideEntryPoints,
 };
 use crate::strip_generics::StripGenerics;
 use crate::utils::{extract_return_type, filter_wheres, process_fields};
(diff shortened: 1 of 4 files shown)


hashedone (commented):
The `EnumMessage::new()` is overgrown, it might be an excellent place to slice it.

Additionally, a style comment to make things more readable. In general, looks good, but it looks like it goes in a bit lazy direction.

sylvia-derive/src/message.rs
```diff
@@ -209,17 +211,18 @@ impl<'a> EnumMessage<'a> {
...
+        let associated_query = parse_associated_custom_type(source, "QueryC");
 
-        let resp_type = match associated_exec {
-            Some(exec) if !custom.has_msg() => exec,
-            _ => custom.msg(),
-        };
+        let resp_type = custom
+            .msg()
+            .or(associated_exec)
+            .unwrap_or_else(Custom::default_type);
```
hashedone, line 220: I am wondering (as it is always `Empty` it could be maybe `Custom::empty`), but it is fine (just thinking about naming). Just for consideration, not a change request.

sylvia-derive/src/message.rs
```diff
@@ -421,9 +424,9 @@ impl<'a> ContractEnumMessage<'a> {
         let variants_constructors = variants.iter().map(MsgVariant::emit_variants_constructors);
         let variants = variants.iter().map(MsgVariant::emit);
 
-        let ctx_type = msg_ty.emit_ctx_type(&custom.query());
+        let ctx_type = msg_ty.emit_ctx_type(&custom.query_or_default());
```
hashedone, line 427: Same here, thinking of `query_or_empty` - an idea, not sure about that. Yes, I know I suggested it before.

hashedone (approved):
LGTM

PR #214, opened 2023-08-10
Cast deps to empty

### LINKED ISSUE #201: Implement the #[sv::custom(query=CustomQuery)] attribute on the contract level

jawoznia:
closes #201


sylvia-derive/src/message.rs
```diff
@@ -901,15 +901,28 @@ impl<'a> GlueMessage<'a> {
             let ContractMessageAttr {
                 variant,
                 has_custom_msg,
+                has_custom_query,
```
hashedone, line 904: I am wondering if we could merge those `has_custom_*` in single structure like:

```rust
struct Customs {
  has_msg: bool,
  has_query: bool,
}
```

Two fields are not much, but I am wondering if it would grow in the future. For now, please don't change it, but things to keep in mind when this grows too much.

sylvia-derive/src/parser.rs
```diff
@@ -262,6 +263,35 @@ fn parse_generics(content: &ParseBuffer) -> Result<Vec<Path>> {
     Ok(params)
 }
 
+fn interface_has_custom(content: ParseStream) -> Result<(bool, bool)> {
```
hashedone, line 266: Going with the `Customs` I previously mentioned would make this signature far more readable - maybe it is actually a good idea already?

sylvia/tests/custom_query.rs
```diff
@@ -134,10 +134,48 @@ mod associated_type_interface {
...
+        fn default_exec(&self, ctx: ExecCtx) -> StdResult<Response>;
+    }
+
+    #[contract(module=super)]
+    #[sv::custom(query=MyQuery)]
+    impl DefaultQueryInterface for crate::MyContract {
+        type Error = StdError;
+
+        #[msg(query)]
+        fn default_query(&self, _ctx: QueryCtx) -> StdResult<SomeResponse> {
```
hashedone, line 163: We need an additional test here.

We need a query variant here with the custom query - taking the `QueryCtx<MyQuery>` context.  Then, in one of the tests, we need to create a multitest test with a Custom module queried by this contract, making sure that the querying works correctly here.

hashedone (commented):
Implementation seems correct, but the critical test is missing.

Also, one styling comment - you decide if you want to follow it.

hashedone (approved):
LGTM

PR #218, opened 2023-09-04
Test custom query

jawoznia:
Add missing `custom_query` parameter in `entry_points` macro expansion.

Example with custom module.


examples/contracts/custom/src/multitest/tests.rs
```diff
@@ -0,0 +1,28 @@
...
+
+#[test]
+fn test_custom() {
+    let owner = "owner";
+
+    let mt_app = cw_multi_test::BasicAppBuilder::<CounterMsg, CounterQuery>::new_custom()
+        .with_custom(CustomModule::new())
+        .build(|router, _, storage| {
+            router.custom.save_counter(storage, 0).unwrap();
+        });
```
hashedone, line 16: It is so good. We need some better way of custom initialization of the modules (customs are not so often, but still having in mind balances), but in general - I like what I see.

What I wonder - do you need the fat-fish syntax? I am not sure how is type elision going there exactly, but it is worth checking - I can imagine that if all the functions are properly generic here, the custom query and messages should be possible to deduce from the `Module` implementation on `CustomModule` - however I am aware that MT generic are a little bit crazy - so compiler might stuck there. If that is a case, I am ok with that.

Also - we should probably add functionality to build this App with Sylvia App/Builder directly (where we could also fix generics) - but not as a part for this PR.

hashedone (approved):
LGTM.

I was worried that custom modules might be troubling, but it seems it is easy enough to work with them. Ready to merge for me.

examples/contracts/custom/src/multitest/tests.rs
jawoznia: Ye, turbofish was not needed here.
This functionality should definitely be added to the `sylvia::App` in later PR.

PR #222, opened 2023-09-18
feat: Generate migrate entry point if message defined on contract

jawoznia:


hashedone (approved):
LGTM, but I would like to see tests for that.

The easiest way to verify that it is working is to add in one of the tests something like:

```
use entry_points::{instantiate, execute, query, migrate, reply};
```

You would do that for a single test that emits entry points. I checked manually they are correctly generated, but adding the verification wouldn't hurt.

jawoznia:
> LGTM, but I would like to see tests for that.
> 
> The easiest way to verify that it is working is to add in one of the tests something like:
> 
> ```
> use entry_points::{instantiate, execute, query, migrate, reply};
> ```
> 
> You would do that for a single test that emits entry points. I checked manually they are correctly generated, but adding the verification wouldn't hurt.

Agree, but I think better way is to create `cw_multi_test::ContractWrapper`. This way we will also check if structure of these entry points is correct.

Issue #226, opened 2023-09-18
Allow implementing generic interface on contract


jawoznia:
This should be done using `#[messages(my::module: query<Param: my + bonds>, exec<Msg: my + bonds>)]`.

PR #237, opened 2023-10-06
Support generic interface on contract

jawoznia:
Part 1/2 of #226

This change lacks `#[messages()]` attribute on main `#[contract]` call as it is already a big PR

### DIFF
--- a/sylvia-derive/src/check_generics.rs
+++ b/sylvia-derive/src/check_generics.rs
@@ -1,26 +1,57 @@
 use syn::visit::Visit;
-use syn::GenericParam;
+use syn::{parse_quote, GenericArgument, GenericParam, Type};
 
-#[derive(Debug)]
-pub struct CheckGenerics<'g> {
-    generics: &'g [&'g GenericParam],
-    used: Vec<&'g GenericParam>,
+pub trait GetPath {
+    fn get_path(&self) -> Option<syn::Path>;
 }
 
-impl<'g> CheckGenerics<'g> {
-    pub fn new(generics: &'g [&'g GenericParam]) -> Self {
+impl GetPath for GenericParam {
+    fn get_path(&self) -> Option<syn::Path> {
+        match self {
+            GenericParam::Type(ty) => {
+                let ident = &ty.ident;
+                Some(parse_quote! { #ident })
+            }
+            _ => None,
+        }
+    }
+}
+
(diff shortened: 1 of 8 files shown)


sylvia-derive/src/check_generics.rs
```diff
@@ -1,26 +1,51 @@
 use syn::visit::Visit;
-use syn::GenericParam;
+use syn::{GenericArgument, GenericParam, Type};
+
+pub trait AsIdent {
```
hashedone, line 4: I really struggle with this. How is there a situation when you want to use `GenericParam` and `GenericArgument` exchangeably? What is even more worrying - in the case of `GenericArgument` implementation disregards any non-path argument, and for paths - it ignores most of the path.

hashedone (commented):
Looks okish, but I really struggle with this `AsIdent` and dropping the entire path for generics verification - it seems to assume that if paths have the same ident, they are the same, which is not the case. We agreed on the call you would double-check this. 

The tests that are the most important look good - I'll check the code generated there as well, but I need to understand what happens with those paths first.

hashedone:
Also, what I noticed in the generated code (I cannot find it in the review, but it is important) - on dispatching `_Phantom` pattern, you make it `unreachable!()`. This doesn't seem right. Messages are external input, and you can get anything there. It can be custom JSON designed for this variant. Deserialization of this message is derived, so it is possible to happen. It should cause an error to be trackable.

Also - it is included in the schema, which needs to be corrected. We need to make PR to the schema generator, which adds syntax for ignoring the variant in the schema. Possible `#[serde(skip)]` would work here - it would also solve my previous comment, and `unreachable!` would be acceptable then (still, I think not preferable). But we need to ensure that schema properly interprets this `#[serde(skip)]`, especially for queries.

hashedone:
Yet another thing I am not sure where is generated - I see that generic types for messages are skipping generics not used in messages. It is messy if someone wants to build a message manually which might be sometimes needed - one would need to know which generics are to be passed. The idea of `_Phantom` variant here is to use generics that are not used in the particular message (eg for `Cw1QueryMsg` I imagine `_Phantom` variant to keep the `Msg` parameter). It is ok for it to keep all of them for simpler implementation.

hashedone (commented):
LGTM, with one test request.

hashedone (approved):

Issue #227, opened 2023-09-18
Support generic contracts


jawoznia:
Messages and helpers generated with `contract` macro should consider generics defined on `impl` block.
Generics from Contract and Interface should be aggregated.

PR #240, opened 2023-10-13
feat: Support generic contract for simple contract

jawoznia:
Part of #227 

This change allows user to create simple Contract with generic `instantiate`.

Missing support for:
- exec, query, migrate, reply generic messages
- impl generic interface on generic contract
- impl non-generic interface on generic contract

### DIFF
--- a/sylvia/tests/generics.rs
+++ b/sylvia/tests/generics.rs
@@ -41,7 +41,7 @@ pub mod whitelist {
     pub trait Whitelist<Msg, QueryRet>
     where
         for<'msg_de> Msg: CustomMsg + Deserialize<'msg_de>,
-        for<'msg_de> QueryRet: CustomQuery + DeserializeOwned,
+        QueryRet: CustomQuery + DeserializeOwned,
     {
         type Error: From<StdError>;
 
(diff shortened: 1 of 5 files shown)


sylvia/tests/generics.rs
```diff
@@ -29,23 +29,166 @@ pub mod cw1 {
...
+    use serde::de::DeserializeOwned;
+    use serde::Deserialize;
+    use sylvia::types::{ExecCtx, QueryCtx};
+    use sylvia_derive::interface;
+
+    #[interface(module=msg)]
+    pub trait Whitelist<Msg, QueryRet>
+    where
+        for<'msg_de> Msg: CustomMsg + Deserialize<'msg_de>,
+        for<'msg_de> QueryRet: CustomQuery + DeserializeOwned,
```
hashedone, line 44: I think this `<'msg_de>` is not needed. Or maybe you forgot `+ 'msg_de`?

hashedone (approved):
LGTM, with a small comment I believe you will fix

sylvia/tests/generics.rs
jawoznia: it's not needed

PR #241, opened 2023-10-16
chore: Remove `MessageSignature` from multitest

jawoznia:


hashedone:
What is this change purpose?

jawoznia:
> What is this change purpose?

This change is not connected directly to the feature but I found it troublesome at some points to work with the original design. I decided to smuggle this small refactor in this feature to ease the further development and to rely on it in future refactor.

Before merge to main I will squash this commit to not pollute commit history with `chore`.

hashedone (approved):
OK, Then LGTM.
Thank you for doing it in its separate PR. Appreciate it.

PR #242, opened 2023-10-20
feat: Support generic exec

jawoznia:


sylvia-derive/src/message.rs
```diff
@@ -1333,21 +1351,23 @@ impl<'a> GlueMessage<'a> {
...
-                    pub fn dispatch #unused_generics #where_clause (
+                impl #bracketed_used_generics #contract_enum_name #bracketed_used_generics {
+                    pub fn dispatch #unused_generics (
                         self,
                         contract: &#contract,
                         ctx: #ctx_type,
-                    ) -> #ret_type {
-                        const _: () = {
+                    ) -> #ret_type #full_where_clause {
+                        const fn assert_no_intersection #bracketed_used_generics () #where_clause {
```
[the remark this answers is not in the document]
jawoznia: ye, it's not a compile time check anymore

sylvia-derive/src/input.rs
```diff
@@ -213,23 +213,13 @@ impl<'a> ImplInput<'a> {
...
-        );
-        let query = self.emit_glue_msg(
-            &Ident::new("QueryMsg", Span::mixed_site()),
-            MsgType::Query,
-            variants,
-        );
+        let exec_impl = self.emit_enum_msg(MsgType::Exec);
+        let query_impl = self.emit_enum_msg(MsgType::Query);
+        let exec = self.emit_glue_msg(MsgType::Exec);
+        let query = self.emit_glue_msg(MsgType::Query);
```
hashedone, line 222: I like how this got reduced

sylvia-derive/src/message.rs
```diff
@@ -1333,21 +1352,23 @@ impl<'a> GlueMessage<'a> {
...
                             let msgs: [&[&str]; #variants_cnt] = [#(#messages_call),*];
                             #sylvia ::utils::assert_no_intersection(msgs);
-                        };
+                        }
+
+                        assert_no_intersection #prefixed_used_generics ();
```
hashedone, line 1371: But if think it can be compile time. You made a function constant; why not do this:
```suggestion
                        const _: () = assert_no_intersection #prefixed_used_generics ();
```
In my opinion, it should force executing `assert_no_intersection` in the compile time.

hashedone (approved):
Looks mostly good to me with one relevant comment - I thnk it still can be a CT check with a minimal change.

sylvia-derive/src/message.rs
jawoznia: Unfortunately this causes the problem I originally encountered:

```
use of generic parameter from outer function type parameter from outer function
```

jawoznia: Solution I went with is to move the `messages` method out of f.e. `ExecMsg` and have it as a separate function named `execute_messages` or `query_messages` to avoid name collissions.

PR #243, opened 2023-10-23
test: Move generic contracts and interfaces to examples

jawoznia:
As generic tests are big and rely on multiple configurations of interfaces and contracts I think it's a good idea to move them to `examples`.

It should allow us also to automate schema generation for those contracts.


examples/contracts/generic_contract/src/contract.rs
```diff
@@ -0,0 +1,99 @@
...
+
+#[contract]
+impl<InstantiateParam, ExecParam, QueryParam, MigrateParam, RetType>
+    GenericContract<InstantiateParam, ExecParam, QueryParam, MigrateParam, RetType>
+where
+    for<'msg_de> InstantiateParam: CustomMsg + Deserialize<'msg_de> + 'msg_de,
+    for<'exec> ExecParam: CustomMsg + DeserializeOwned + 'exec,
+    for<'exec> QueryParam: CustomMsg + DeserializeOwned + 'exec,
+    for<'exec> MigrateParam: CustomMsg + DeserializeOwned + 'exec,
+    for<'ret> RetType: CustomMsg + DeserializeOwned + 'ret,
```
hashedone, line 30: I am wondering - shouldn't we go with `+ 'static` instead of a generalized lifetime? Don't really see when you want to use non-static type here.

examples/contracts/generic_contract/src/contract.rs
```diff
@@ -0,0 +1,99 @@
...
+        use super::multitest_utils::CodeId;
+        let app = App::default();
+        let code_id: CodeId<
+            cw_multi_test::BasicApp<Empty, Empty>,
+            ExternalMsg,
+            ExternalMsg,
+            ExternalMsg,
+            super::ExternalMsg,
+            super::ExternalMsg,
+        > = CodeId::store_code(&app);
```
hashedone, line 81: Do we really need to pass all those generics? Especially the `BaxicApp` - cannot it be `_`? The rest are fairly okay as those are just contract types.
Also, I think the `App` should be the last generic - so the user can completely ignore it as it should be elideable.

hashedone (approved):
Mostly LGTM, with two small comments to address

examples/contracts/generic_contract/src/contract.rs
jawoznia: Not sure but I decided to go with custom lifetime to cover expanded predicate. I can change three of them to static

examples/contracts/generic_contract/src/contract.rs
jawoznia: Unfortunately they are required as `CodeId` is generic over types defined on contract.

We can omit the `App` though and I think it's a good idea to move it to last argument.

PR #247, opened 2023-10-25
feat: Wrap contract types in `sv` module

### LINKED ISSUE #114: Consider better structure of generated code
Right now, generated code is bloated into the namespace where macros are used. There are also strange limitations on naming of the modules for traits, which creates a place for mistakes.

I would prefer to isolate things a bit. All generated things should go to `sv` module (so if we `#[contract]`, it would have things generated as `sv::ExecMsg`, `sv::QueryMsg`, etc). This way, rules about how to avoid naming collisions would be easy: single contract/interface call per module, but we eventually want the module to be named however the user wants to. Additionally "if you want to use sylvia, do not create `sv` modules yourself).

Additionally, I want to rename all tests-related generated modules to `mt` - we want to keep them short to be easy to type around.

jawoznia:
close #114

### DIFF
--- a/examples/contracts/custom/src/bin/schema.rs
+++ b/examples/contracts/custom/src/bin/schema.rs
@@ -1,6 +1,6 @@
 use cosmwasm_schema::write_api;
 
-use custom::contract::{ContractExecMsg, ContractQueryMsg, InstantiateMsg};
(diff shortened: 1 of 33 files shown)


hashedone (approved):
LGTM

Issue #86, opened 2023-03-14
Introduce the extension traits for the contract type


hashedone:
We are getting to the point where many things are generated, and there is more than one macro invocation in a single crate. It creates an issue of naming collisions when entities are generated in a single module. It could be solved by forcing devs to have only one macro invocation per module. Still, it is significantly limiting in terms of contract structure, and it is beneficial to leave as much freedom as possible to the developer. 

The other idea is to keep generated entities in some enclosed scope and then expose some traits on contracts that access generated entities instead of modules. This makes the naming problem easier. It also vastly improves documentation easiness and might make LSP work better.

What I imagine is, for example, the `sylvia::Contract` trait looking like this:

```rust
trait Contract {
  type InstantiateMsg;
  type ExecMsg;
  type QueryMsg;
  type MigrationMsg;
}
```

And then, there is implementation generated for the contract:

```rust
impl Contract for Cw20Contract {
  type InstantiateMsg = msg::InstantiateMsg;
  type ExecMsg = msg::ExecMsg;
  type QueryMsg = msg::QueryMsg;
  type MigrationMsg = sylvia::cw_std::Empty;
}
```

It could be tempting to use this trait also to add all other meta-info, let say:

```rust
trait Contract {
  type InstantiateMsg;
  type ExecMsg;
  type QueryMsg;
  type MigrationMsg;
  
  type MtHelper = multi_test::Cw20Contract;
  type MtCodeId = multi_test::CodeId;
}
```

But then, the trait implementation would have to be split between macros, which seems impossible. Also, sometimes some entities need to be generated, which would complicate such a case. The solution is to have more crates:

```rust
trait Multitest {
  type CodeId = cw20::multi_test::CodeId;
  type Contract = cw20::multi_test::Contract;
}
```

Then instead of accessing generated entities directly, one imports the proper trait and uses the contract type:

```rust
let code_id = Cw20Contract::CodeId::store(&App);
```

This would make finding and accessing generated entities easier, as you don't need to remember Sylvia naming patterns. The next step is to add the `sylvia::prelude` module with all those traits, so it is easier to import everything at once with a prelude pattern.

The final step would be to hide generated implementations using generated namespace, which the user will not likely use. This would free the names generated by Sylvia for dev, as they are meant to be accessed by exposed traits, which are very quickly findable.

jawoznia:
[...]

PR #248, opened 2023-10-26
Contract api

jawoznia:
Impl Extension trait on Contract. This change is required to generate `entry_point`s without requiring user to specify which generics are for which messages.

Unfortunately rust compiler requires fully-qualified path and in current state it's not the most user friendly Api so the primary use case is code generation.

#86

### DIFF
--- a/sylvia/tests/api.rs
+++ b/sylvia/tests/api.rs
@@ -1,0 +1,127 @@
+use cosmwasm_std::{Response, StdResult};
+use std::marker::PhantomData;
+
+use sylvia::types::{CustomMsg, ExecCtx, InstantiateCtx, MigrateCtx, QueryCtx};
+use sylvia_derive::contract;
+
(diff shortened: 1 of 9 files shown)


sylvia/tests/api.rs
```diff
@@ -0,0 +1,135 @@
...
+        // let _: crate::some_contract::sv::BoundQuerier<'_, cosmwasm_std::Empty> = <SomeContract<
+        //     SvCustomMsg,
+        //     SvCustomMsg,
+        //     SvCustomMsg,
+        //     SvCustomMsg,
+        //     SvCustomMsg,
+        // > as ContractApi>::Querier::borrowed(
+        //     &owner,
+        //     &querier_wrapper,
+        // );
```
jawoznia, line 133: Not sure why compiler expects `querier_wrapper` and `owner` to be `'static` in this test.

hashedone (approved):
LGTM

PR #249, opened 2023-10-26
feat: Support generic types in entry points

### LINKED ISSUE #228: Use generics in entry_points macro
`entry_points` require solid types. We have to parse them as attributes in `entry_points<String>` and generate entry points with `Contract` generic over these types.

jawoznia:
close #228


hashedone (approved):
LGTM

PR #250, opened 2023-10-27
fix: Expect `,` in `:custom(msg, query)`

jawoznia:
This check was missed in tests. Found it while updating the sylvia book.


hashedone (approved):
Uhh, so before there was no comma between `query=...` and `msg...`? Good catch, good it's fixed.

jawoznia:
> Uhh, so before there was no comma between `query=...` and `msg...`? Good catch, good it's fixed.

Unfortunately. I missed this check earlier.

PR #251, opened 2023-11-07
tests: Support generic field in contract

### LINKED ISSUE #227: Support generic contracts

jawoznia:
close #227


jawoznia:
closing as it is already being delivered under #249

PR #256, opened 2023-11-15
fix: Allow generic return type in contract queries

jawoznia:


hashedone (commented):
I am not really sure which test is testing the generic ret type in queries, what do I miss?

jawoznia:
closing in favor of #264

Issue #252, opened 2023-11-09
Sudo support


jawoznia:
# Motivation

Since `Sylvia` supports `CustomMsg` and `CustomQuery` we should add support for `sudo`.

# Generate `SudoMsg`

To allow `sudo` communication with a contract, first we have to generate the `SudoMsg`.
It should be represented as and `enum` and concatenate messages as variants.
`Sudo` should be supported by both `interface` and `contract` and generated as

```rust
pub enum SudoMsg {
    MoveFunds {
        recipient: String,
        amount: Vec<Coin>,
    }
}
```

## `Sudo` in Interface

First step to provide support for the sudo would be to add support for it in interfaces.
We should support below example and generate `SudoMsg` with two variants.

```rust
#[interface]
pub trait MyInterface {
    type Error: From<StdError>;

    #[msg(sudo)]
    fn move_funds(&self, ctx: SudoCtx, recipient: String, funds: Vec<Coin>) -> Result<Response, Self::Error>;

    #[msg(sudo)]
    fn send_funds(&self, ctx: SudoCtx, recipient: String, funds: Vec<Coin>) -> Result<Response, Self::Error>;
}
```

This should look the same as `ExecMsg` and `QueyMsg` so not much to be covered here.

## `Sudo` in Contract

Same as in case of the Interface `SudoMsg` should be generated as `ExecMsg` and `QueryMsg`.
There is however one additional thing to be generated which is `ContractSudoMsg`.
It is needed as we have to be able to implement Interfaces on the Contracts and we want to be able to dispatch to proper handler.

Next to the generated messages we have to generate utils for handling the `sudo` in the `MultiTest` helpers. TODO!

# `Sudo` in Interface Implemenation

Interface implementation handles two things:
- `interface::Querier` implementation on `contract::BoundQuerier`
- newly generated trait implementation on `interface::Proxy`

We don't have to update `Querier` generation, however `sudo` messages should be implemented on the `interface::Proxy`.
At this point we should be able to have a Contract handling `sudo` messages from both itself and Interface and be able to test the behavior using `MultiTest`.

# Update docs

`README` and `sylvia-book` has to be updated with changes from this feature.

jawoznia:
close #139

jawoznia:
closed with #317

PR #257, opened 2023-11-15
feat: Generate SudoMsg in contract

jawoznia:
#252

### DIFF
--- a/sylvia-derive/src/message.rs
+++ b/sylvia-derive/src/message.rs
@@ -2,8 +2,8 @@ use crate::check_generics::{CheckGenerics, GetPath};
 use crate::crate_module;
 use crate::interfaces::Interfaces;
 use crate::parser::{
-    parse_associated_custom_type, parse_struct_message, ContractErrorAttr, ContractMessageAttr,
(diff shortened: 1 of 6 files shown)


sylvia-derive/src/message.rs
```diff
@@ -1290,32 +1290,7 @@ impl<'a> GlueMessage<'a> {
...
-            match (msg_ty, customs.has_msg) {
-                (MsgType::Exec, true) => quote! {
-                    #contract_enum_name:: #variant(msg) => #sylvia ::into_response::IntoResponse::into_response(msg.dispatch(contract, Into::into( #ctx ))?)
-                },
-                _ => quote! {
-                    #contract_enum_name :: #variant(msg) => msg.dispatch(contract, Into::into( #ctx ))
-                },
-            }
-        });
+        let dispatch_arms = interfaces.emit_dispatch_arms(msg_ty);
```
hashedone, line 1293: Ideally, such refactoring should be one in separate PR. It could be accepted immediately, and this one would be simpler.

hashedone (commented):
LGTM

sylvia-derive/src/message.rs
jawoznia: you are right

PR #259, opened 2023-11-16
feat: Allow specifying salt for contract address

jawoznia:


sylvia/Cargo.toml
```diff
@@ -24,12 +31,17 @@ serde = { version = "1.0", default-features = false, features = ["derive"] }
 serde-cw-value = "0.7.0"
 serde-json-wasm = "1.0.0"
 konst = "0.3"
-cw-multi-test = { version = "0.18.0", optional = true }
+cw-multi-test = { version = "0.18.0", optional = true, features = [
+    "cosmwasm_1_2",
```
hashedone, line 35: You cannot enable "cosmwasm_1_2" in mt and not enable it in cosmwasm-std. You are asking yourself for problems. You need to have a `cosmwasm_1_2` feature flag, which enables those two. Then everything that needs those flags to work, should be behind it (we can leave setting salt itself, but ignore it when instantiating without this flag).

hashedone (commented):
Feature management has to improve

sylvia-derive/src/multitest.rs
```diff
@@ -453,6 +453,48 @@ where
...
+                            _phantom: std::marker::PhantomData::default(),
+                        })
+                }
+            }
+        } else {
+            #[cfg(not(tarpaulin_include))]
+            {
+                quote! {
+                    let err = #sylvia ::cw_std::StdError::generic_err(
+                        "`with_salt` was called, but it requires `cosmwasm_1_2` feature enabled. Consider removing `with_salt` or adding the `cosmwasm_1_2` feature."
```
hashedone, line 491: I am not sure about making it an error. If it were possible to emit it as a warning, it would be better - something like "`with_salt` requires `cosmwasm_1_2` to have an effect. The salt would be ignored; consider enabling `cosmwasm_1_2` if you want predictable address initialization support".

Then in the case of missing `cosmwasm_1_2` I would fall back to `instantiate` and ignore salt.

hashedone (commented):
Mostly ok, but consider falling back to instantiate instead of hard error when using salt pre-1.2

hashedone (approved):

sylvia-derive/src/multitest.rs
jawoznia: My idea was that `with_salt` does nothing if we don't enable the `cosmwasm_1_2` feature.
If called without the feature user most likely want `insantiate2` to be called but it never will.
There is no reason to keep this call in such case and should either be removed or feature should be enabled.

But maybe the error is actually a better solution here.

hashedone: Ok, that technically works as well. I just think its slightly more maintenance added. But I don't really mind about that.

PR #260, opened 2023-11-16
feat: Allow querying code_info from CodeId and App

jawoznia:


sylvia-derive/src/multitest.rs
```diff
@@ -488,6 +488,10 @@ where
                         self.code_id
                     }
 
+                    pub fn code_info(&self) -> #sylvia ::cw_std::StdResult< #sylvia ::cw_std::CodeInfoResponse> {
```
hashedone, line 491: This should be behind feature flag as well (see the other review about instantiate2)

hashedone (commented):
Feature management

hashedone (approved):
LTGM

Issue #263, opened 2023-11-21
Generics in interface implementation


jawoznia:
With version `0.9.0` users are able to define generics in their contracts and interfaces.
Although contracts generics can be set as any concrete types by users importing these contracts to their own it is impossible to forward these concrete types to the implemented interfaces.

# Complexity

This functionality has to cover a lot of edge cases and will not be easy to implemented.

## Generics passed to the interface implemented on non-generic contract

This should be fairly simple as there is no intersection between generics from the interface and the contract.

```rust
#[contract]
impl<ExecT, QueryT> Generic<ExecT, QueryT> for NonGeneric {}
```

## Concrete types passed to the interface implemented on non-generic contract

At this stage we would have to detect which generic arguments are also generic parameters defined in `impl` block. Mechanism for that should be fairly simple, but good coverage is required to detect every place it should be appllied.

```rust
#[contract]
impl Generic<SvCustomMsg, SvCustomMsg> for NonGeneric {}
```

## Generic types passed to the interface implemented on generic contract

Here generics might overlap in generated code. We have to filter the duplications. `GlueMessage` will for sure require change as it generates the wrapper message for `exec` and `query`.

```rust
#[contract]
impl<T1, T2, T3> Generic<T1, T2> for Generic<T1, T2, T3> {}
```

## Concrete types passed to the interface implemented on generic contract

Same as in case of non-generic contract we have to detect the concrete types and not pass them to generated types as generics. 

```rust
#[contract]
impl<T1, T2, T3> Generic<SvCustomMsg, SvCustomMsg> for Generic<T1, T2, T3> {}
```

## Separate generic types passed to the interface

Some types passed to the interface might be generics and some concrete types. We should filter out the concrete types from the generics.

Behavior differ in `execs` and `queries` and even in parameters and return types and we have to cover all of these cases.

```rust
#[contract]
impl<T1, T2, T3> Generic<T1, SvCustomMsg> for Generic<T1, T2, T3> {}
```

# Implemantation

As we base on existing examples I think we can apply these changes on a feature branch step by step by disabling interfaces implementations and uncommenting them PR by PR.

hashedone:
Closing in favor of proper Assoc Types in interfaces - generic interfaces will not support.

PR #264, opened 2023-11-21
feat: Forward generics to the interface

jawoznia:
Part of #263

### DIFF
--- a/examples/contracts/generic_contract/src/generic.rs
+++ b/examples/contracts/generic_contract/src/generic.rs
@@ -8,8 +8,7 @@ use sylvia::types::{ExecCtx, QueryCtx, SvCustomMsg};
 #[contract(module = crate::contract)]
 #[messages(generic as Generic)]
 #[sv::custom(msg=SvCustomMsg, query=SvCustomQuery)]
-impl<InstantiateT, ExecT, QueryT, MigrateT, FieldT>
-    Generic<SvCustomMsg, SvCustomMsg, sylvia::types::SvCustomMsg>
+impl<InstantiateT, ExecT, QueryT, MigrateT, FieldT> Generic<SvCustomMsg, SvCustomMsg, SvCustomMsg>
     for crate::contract::GenericContract<InstantiateT, ExecT, QueryT, MigrateT, FieldT>
 where
     for<'msg_de> InstantiateT: CustomMsg + Deserialize<'msg_de> + 'msg_de,
@@ -38,11 +37,7 @@ where
     // F.e. if we this query would return `SvCustomMsg` and we would pass
     // `sylvia::types::SvCustomMsg` to the `Generic` trait paths would not match.
     #[msg(query)]
-    fn generic_query(
-        &self,
-        _ctx: QueryCtx,
-        _msg: SvCustomMsg,
-    ) -> StdResult<sylvia::types::SvCustomMsg> {
+    fn generic_query(&self, _ctx: QueryCtx, _msg: SvCustomMsg) -> StdResult<SvCustomMsg> {
         Ok(SvCustomMsg {})
     }
 }
--- a/examples/contracts/generics_forwarded/src/generic.rs
+++ b/examples/contracts/generics_forwarded/src/generic.rs
@@ -3,29 +3,35 @@ use generic::Generic;
 use serde::de::DeserializeOwned;
 use serde::Deserialize;
 use sylvia::contract;
-use sylvia::types::{ExecCtx, QueryCtx, SvCustomMsg};
+use sylvia::types::{CustomQuery, ExecCtx, QueryCtx, SvCustomMsg};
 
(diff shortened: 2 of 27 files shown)


examples/contracts/generic_contract/src/generic.rs
```diff
@@ -32,12 +33,16 @@ impl<InstantiateParam, ExecParam, QueryParam, MigrateParam, RetType, FieldType>
     // It's because we have to map unique generics used as they can be used multiple times.
     // If for some reason like here one type would be used in place of two generics either full
     // path or some alias has to be used.
+    //
+    // Sylvia will fail to recognize generic used if their path is different.
+    // F.e. if we this query would return `SvCustomMsg` and we would pass
+    // `sylvia::types::SvCustomMsg` to the `Generic` trait paths would not match.
```
hashedone, line 39: I don't understand it, and I don't like this limitation. It should work as in Rust. Could you explain what does it exactly mean?

examples/contracts/generics_forwarded/src/generic.rs
```diff
@@ -0,0 +1,89 @@
...
+    FieldT: 'static,
+{
+    type Error = StdError;
+
+    #[msg(exec)]
+    fn generic_exec(&self, _ctx: ExecCtx, _msgs: Vec<CosmosMsg<ExecT>>) -> StdResult<Response> {
+        Ok(Response::new())
+    }
+
+    // Sylvia will fail if single type is used to match against two different generics
```
hashedone, line 38: Wait, that means that I can't have both generics being monomorphized to `String`? That cannot be the case, and using particular solutions is not an excuse.

examples/contracts/generics_forwarded/src/generic.rs
```diff
@@ -0,0 +1,89 @@
...
+    // It's because we have to map unique generics used as they can be used multiple times.
+    // If for some reason like here one type would be used in place of two generics either full
```
hashedone, line 40: There are plenty of reasons for that, and we cannot make people using aliases for every generic String.

examples/contracts/generics_forwarded/src/generic.rs
```diff
@@ -0,0 +1,89 @@
...
+    // path or some alias has to be used.
+    //
+    // Sylvia will fail to recognize generic used if their path is different.
+    // F.e. if we this query would return `SvCustomMsg` and we would pass
+    // `sylvia::types::SvCustomMsg` to the `Generic` trait paths would not match.
```
hashedone, line 45: Again - I don't understand it, but it scares me.

hashedone (commented):
ATM, I don't like examples. Limitations are unacceptable really. If you have no idea how to work them around, I can take a closer look at implementation, but I didn't really use much time there as the API doesn't work for me.

examples/contracts/generic_contract/src/generic.rs
jawoznia: This is because when we check if generic type is used in context of given type it compares their full `path`.
In such case `Msg`, `types::Msg` and `sylvia::types::Msg` are three different types.

examples/contracts/generics_forwarded/src/generic.rs
jawoznia: These comments are actually related to the `generic_contract` where we pass the concrete types to the generics.
Comment landed here when I copied the examples.

examples/contracts/generics_forwarded/src/generic.rs
jawoznia: These comments are actually related to the `generic_contract` where we pass the concrete types to the generics.
Comment landed here when I copied the examples.

examples/contracts/generic_contract/src/generic.rs
hashedone: I don't find this explains everything. `types::Msg` and `sylvia::types::Msg` are obviously not generic parameters - they are coming from a module, so you just should assume they exist and let Rust resolve it. `Msg` is either a generic type or a type available in global scope. Still, its Rust that would handle figuring out the exact type, but for additional codegen you should consider a type being maybe a generic argument only if it has no path besides the ident. If that is the case, you should be able to see the generic parameters of your definition block. If the ident you face is defined as generic, it is generic. Otherwise, it is not generic, and you don't make any further assumptions about that. I need help understanding how scoping makes any deal here. Could you give some very particular examples of what is possible and what is not? And I am not asking about a description of an example, but an example itself - I think some misunderstanding is happening here. We can a call about this if you want as it seems to be not trivial.

hashedone (commented):
Before going further I need to exactly understand those limitations. Let's setup a call about that.

hashedone (approved):
Mostly good. Unresolved comments to be handled under https://github.com/CosmWasm/sylvia/issues/263

PR #268, opened 2023-11-24
Impl sudo interface

jawoznia:
part of #252 

Simple example in `sylvia/tests` confirming that sudo interface can be implemented on contract

### DIFF
--- a/sylvia/tests/dispatching.rs
+++ b/sylvia/tests/dispatching.rs
@@ -19,7 +19,7 @@ pub struct QueryResponse {
 mod interface {
     use cosmwasm_std::{Addr, Decimal, Response, StdError};
     use sylvia::interface;
-    use sylvia::types::{ExecCtx, QueryCtx};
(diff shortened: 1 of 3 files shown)


sylvia/tests/dispatching.rs
```diff
@@ -108,6 +128,7 @@ mod contract {
     pub struct Contract {
         pub(crate) execs: RefCell<u64>,
         pub(crate) queries: RefCell<u64>,
+        pub(crate) sudos: RefCell<u64>,
```
hashedone, line 131: I didn't notice it before, but it would be much better if those entities would be contract state, and then you can always raw-query them in the test. Could we fix all 4 of those (making data a map) in this PR?

hashedone (commented):
Looks mostly good to me, but I would like to make the contract internals a proper state

sylvia/tests/dispatching.rs
jawoznia: Changed all but `queries` as we don't have access to mutable storage query.

PR #269, opened 2023-11-24
Sudo entry point

jawoznia:
part of #252


hashedone (approved):
LGTM. I like this level of testing.

PR #273, opened 2023-12-01
feat: Support duplicated exec generic params

jawoznia:
part of #263

### DIFF
--- a/sylvia-derive/src/message.rs
+++ b/sylvia-derive/src/message.rs
@@ -906,15 +906,6 @@ where
         &self.unused_generics
     }
 
-    pub fn as_where_clause(&'a self) -> Option<WhereClause> {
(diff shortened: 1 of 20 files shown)


sylvia-derive/src/message.rs
```diff
@@ -830,6 +830,13 @@ impl<'a> MsgVariant<'a> {
     }
 }
 
+// TODO: This could be made msg_type agnostic and stores map of msg_ty/variants.
```
hashedone, line 833: Nope. Remove it from here, and create an issue about that. Noone ever tracked TODOs in code comments.

hashedone (approved):
LGTM in general, I was surprised to see it as I thought we would first fix the general generic trait issues, but happy to accept.
One fix request - please get rid of a TODO comment, and make it into a GH issue.

PR #281, opened 2024-01-10
feat: Interfaces generate with associated_types

jawoznia:


sylvia-derive/src/associated_types.rs
```diff
@@ -0,0 +1,76 @@
...
+                    Some(ty)
+                }
+                _ => None,
+            })
+            .collect();
+
+        Self(associated_types)
+    }
+
+    pub fn as_where_predicates(&self) -> Vec<WherePredicate> {
```
kulikthebird, line 26: Nitpick: Consider returning an iterator

sylvia-derive/src/associated_types.rs
```diff
@@ -0,0 +1,76 @@
...
+                let colon = &associated.colon_token;
+                let bound = &associated.bounds;
+                parse_quote! { #name #colon #bound }
+            })
+            .collect()
+    }
+
+    pub fn as_where_clause(&self) -> Option<WhereClause> {
+        let predicates = self.as_where_predicates();
+        match !predicates.is_empty() {
```
kulikthebird, line 40: Nitpick: match --> if

kulikthebird (changes requested):

sylvia-derive/src/message.rs
```diff
@@ -612,10 +522,35 @@ impl<'a> MsgVariant<'a> {
         }
     }
 
+    pub fn emit_trait_querier_impl(&self, associated_name: &[&Ident]) -> TokenStream {
```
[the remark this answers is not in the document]
jawoznia: This method emits variant declaration and is strictly related to the message variants. As it operates almost entirely on data related to the variants I would leave it here.

We could think about moving the `MsgVariant` somewhere as it's huge on it's own and it's goal is to emit message related methods while other structs like `EnumMessage` are collectors emitting whole messages.

sylvia-derive/src/message.rs
```diff
@@ -787,6 +722,10 @@ impl<'a> MsgVariant<'a> {
         }
     }
 
+    pub fn is_of_type(&self, msg_type: MsgType) -> bool {
```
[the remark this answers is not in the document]
jawoznia: It's removed in further PR so I will leave it here

sylvia-derive/src/associated_types.rs
jawoznia: It is used only in a single place that's why I decided to make it more "specialized"

sylvia-derive/src/associated_types.rs
jawoznia: ye, makes sense. I guess I like match statement too much ^^

kulikthebird (approved):

PR #282, opened 2024-01-11
feat: Impl interface with associated types on non generic contract

jawoznia:


sylvia-derive/src/message.rs
```diff
@@ -1203,6 +1157,27 @@ impl<'a> MsgField<'a> {
         }
     }
 
+    pub fn emit_method_field_with_self<GenericT: GetPath>(
```
jawoznia, line 1160: Consider merging with `emit_method_field`.

By default all of the types should be stripped with `Self` and later we might decide to use `AddSelfPath`.

sylvia-derive/src/remote.rs
```diff
@@ -45,12 +47,17 @@ impl<'a> Remote<'a> {
...
+                        _phantom: std::marker::PhantomData
+                    }
                 }
 
                 pub fn borrowed(addr: &'a #sylvia ::cw_std::Addr) -> Self {
-                    Self{addr: std::borrow::Cow::Borrowed(addr), _phantom: std::marker::PhantomData}
-                }
+                    Self {
+                        addr: std::borrow::Cow::Borrowed(addr),
+                        _phantom: std::marker::PhantomData}
```
jawoznia, line 59: nit: format

sylvia-derive/src/strip_self_path.rs
```diff
@@ -13,3 +15,29 @@ impl Fold for StripSelfPath {
         syn::fold::fold_path(self, Path { segments, ..path })
     }
 }
+
+pub struct AddSelfPath<'a, GenericT>(&'a [&'a GenericT]);
```
kulikthebird, line 19: This is something I'd rather avoid, but as for now I still don't have enouch knowledge about sylvia. I'm happy to once again talk about adding Self to generics and maybe have some brainstorming on other solutions.

sylvia-derive/src/message.rs
```diff
@@ -722,26 +712,25 @@ impl<'a> MsgVariant<'a> {
         }
     }
 
-    pub fn is_of_type(&self, msg_type: MsgType) -> bool {
-        self.msg_type == msg_type
-    }
-
     pub fn emit_proxy_methods_declarations(
```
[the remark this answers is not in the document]
jawoznia: Let's discuss it after 1.0.0 release. I agree we should make the modules more specialized. In this case I think moving the MsgVariants somewhere should do the trick as it's goal is broader then just emit a single message.

sylvia-derive/src/strip_self_path.rs
jawoznia: Agree. We can maybe setup some biweekly or something calls and discuss stuff like that there

sylvia-derive/src/message.rs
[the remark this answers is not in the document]
kulikthebird: This should be covered by https://github.com/CosmWasm/sylvia/issues/295

kulikthebird (approved):

PR #291, opened 2024-01-22
feat: ExecC and QueryC used in place of CustomMsgT

jawoznia:


sylvia-derive/src/associated_types.rs
```diff
@@ -76,6 +90,14 @@ impl<'a> AssociatedTypes<'a> {
...
         }
     }
+
+    pub fn filtered(&self) -> impl Iterator<Item = &&TraitItemType> {
+        self.0.iter().filter(|associated| {
+            !RESERVED_TYPES
+                .iter()
+                .any(|reserved| reserved == &associated.ident.to_string().as_str())
+        })
+    }
```
kulikthebird, line 100:
```suggestion

    pub fn filtered(&self) -> impl Iterator<Item = &TraitItemType> {
        self.0.iter().filter(|associated| {
            !RESERVED_TYPES
                .iter()
                .any(|reserved| reserved == &associated.ident.to_string().as_str())
        }).cloned()
    }
```

sylvia-derive/src/associated_types.rs
```diff
@@ -47,11 +59,13 @@ impl<'a> AssociatedTypes<'a> {
...
     pub fn as_names(&self) -> Vec<&Ident> {
-        self.0.iter().map(|associated| &associated.ident).collect()
+        self.filtered()
+            .map(|associated| &associated.ident)
+            .collect()
     }
 
-    pub fn as_types_declaration(&self) -> &Vec<&TraitItemType> {
-        &self.0
+    pub fn as_types_declaration(&self) -> Vec<&&TraitItemType> {
```
kulikthebird, line 67: Return an iterator (i.e. it would be the same as `filtered()` method) and let decide the user how to operate on this

sylvia-derive/src/associated_types.rs
```diff
@@ -5,7 +5,8 @@ use syn::{
     WhereClause, WherePredicate,
 };
 
-const RESERVED_TYPES: [&str; 3] = ["Error", "QueryC", "ExecC"];
+const ERROR_TYPE: &str = "Error";
+const RESERVED_TYPES: [&str; 3] = [ERROR_TYPE, "QueryC", "ExecC"];
```
kulikthebird, line 9: If there are some reserved type it would be good to describe them in README, examples and at the end in the Sylvia book as well.  It is also relevant whether they are optional or mandatory.

kulikthebird (changes requested):
There are some comments that are worth discussion, but other than that lgtm

kulikthebird (approved):

PR #300, opened 2024-02-02
Update README.md

costa2400:
added a short description with link to Sylvia-template.


jawoznia (approved):
LGTM with one suggestion.
What do you think about merging this section with above line about `sylvia-book`.
I think this could be a good "how to get started" section.

PR #301, opened 2024-02-05
test: Add sudo to custom example

jawoznia:
Part of #252

### DIFF
--- a/examples/contracts/custom/src/multitest/tests.rs
+++ b/examples/contracts/custom/src/multitest/tests.rs
@@ -24,4 +24,9 @@ fn test_custom() {
 
     let count = contract.query_custom().unwrap().count;
     assert_eq!(count, 1);
+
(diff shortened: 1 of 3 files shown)


examples/contracts/custom/src/multitest/tests.rs
```diff
@@ -24,4 +24,9 @@ fn test_custom() {
 
     let count = contract.query_custom().unwrap().count;
     assert_eq!(count, 1);
+
+    contract.sudo_custom().unwrap();
```
kulikthebird, line 28: Is there any test case asserting that when we call sudo_custom, the Sudo message is generated and sent under the hood? I can imagine a situation where `#[msg(sudo)]` and `sudo_custom()` might, by mistake, operate on an Exec or Query message.

jawoznia: tracked under #316

PR #312, opened 2024-02-27
chore: Remove tarpaulin exclusions

kulikthebird:


sylvia-derive/src/message.rs
```diff
@@ -1662,65 +1624,61 @@ impl<'a> EntryPoints<'a> {
...
+                    Some(_) => quote! {},
+                    None => variants.emit_default_entry_point(
+                        &custom_msg,
+                        &custom_query,
+                        name,
+                        error,
+                        &attrs.generics,
+                    ),
+                },
+            );
```
jawoznia, line 1640: We might keep disabling tarpaulin here. Unfortunately it seems that it still has some troubles with coverage inside the iterators.

jawoznia (approved):
Good to see that tarpaulin is finally working. Guess switch `llvm` might did the job.

sylvia-derive/src/message.rs
kulikthebird: As we discussed offline, I'd keep this part as not test-covered so we could research for the solution for the problem in the future.

PR #321, opened 2024-03-05
feat: BoundQuerier improve

kulikthebird:


jawoznia (approved):
LGTM. I very much like this change. Hope we will find out a way to skip `Error = ()` in production code.

PR #322, opened 2024-03-07
feat: Update deps to 2.0.0 (#308)

### LINKED ISSUE #308: Update dependencies
cw, cw-utils, cw2, cw-storage-plus, cw-multi-test: 2.0

jawoznia:
close #308


sylvia/tests/replies.rs
```diff
@@ -77,9 +79,13 @@ mod reply_contract {
 fn entry_point_generation() {
     let msg = Reply {
         id: 0,
+        payload: Default::default(),
+        gas_used: 0,
+        #[allow(deprecated)]
```
kulikthebird, line 84: Is there a way to update this test without adding `allow(deprecated)`?

README.md
```diff
@@ -777,13 +777,13 @@ mod tests {
```
kulikthebird, line 777: Comment to other part of README:

Update the example project's deps (for e.g. `cosmwasm-std = "1.2.5"`) to the proper version. Also let's make sure that this corresponds with the `sylvia-template`

kulikthebird (changes requested):
LGTM, only minor comments

sylvia/tests/replies.rs
jawoznia: I think there is not. There is an issue about it https://github.com/rust-lang/rust/issues/102777, but no action so far.

PR #331, opened 2024-03-21
docs: Update README

kulikthebird:


README.md
```diff
@@ -179,142 +73,100 @@ impl MyContract<'_> {
...
-function argument.
-
-Now let's add an execution message to the contract:
-
-```rust
-#[contract]
-impl MyContract<'_> {
     #[sv::msg(exec)]
-    pub fn increment(&self, ctx: ExecCtx) -> StdResult<Response> {
+    pub fn increment(&self, ctx: ExecCtx) -> Result<Response, ContractError> {
```
jawoznia, line 83: Make it proper sylvia contract. If you switch to different error type add `sv::error` attribute to macro call

README.md
```diff
@@ -179,142 +73,100 @@ impl MyContract<'_> {
...
 error type - it will all be commonized in the generated dispatching function (so
 entry points have to return `ContractError` as its error variant).
 
-Messages equivalent to execution messages are generated.
-Again entry point is already generated like in case of execute and instantiate.
 
 ## Interfaces
 
-One of the fundamental ideas of Sylvia's framework are interfaces, allowing the
+One of the fundamental ideas of Sylvia's framework is interfaces, allowing the
```
jawoznia, line 169: either is interface or are interfaces

jawoznia (approved):
As discussed offline it would be better to pick a single path: verbose or modest README.
I think we should move toward the modest one and make it a showcase what sylvia can do. Then for more info user would have to go and check the `sylvia-book` and/or docs.

In such case I think it's a LGTM step towards clearing redundant information from the README.

Issue #211, opened 2023-08-02
Update code documentation


jawoznia:
If we will go to the https://docs.rs/sylvia/latest/sylvia/#modules we will see "Most of implementation lies in cw-derive-ng crate which is reexported here" which is no longer true. It's probable that there are more missing updates in the docs comments.

Also `sylvia/multitest` types don't appear on docs.rs.

kulikthebird:
This task is already done

PR #336, opened 2024-04-02
docs: Document `sylvia::types` module (#211)

jawoznia:


sylvia/src/types.rs
```diff
@@ -1,23 +1,79 @@
+//! Module providing utilities to build and use sylvia contracts.
 use cosmwasm_std::{Deps, DepsMut, Empty, Env, MessageInfo};
 use schemars::JsonSchema;
 use serde::de::DeserializeOwned;
 use serde::{Deserialize, Serialize};
 
+/// Wrapper around [QuerierWrapper](cosmwasm_std::QuerierWrapper) for more user friendly query experience.
```
kulikthebird, line 7: We might add also an example of calling query method on an interface

kulikthebird (approved):
One extra thing to consider, lgtm.

sylvia/src/types.rs
jawoznia: Good catch. Done

PR #343, opened 2024-04-04
feat: Assert `new` method defined (#342)

### LINKED ISSUE #342: Assert parameterless `new` method is defined
`new` method is required for `sylvia` to dispatch message payload to proper contract method.
It most likely should never accept any parameters as those would have to be received in an entry point which has fixed input.

jawoznia:
closes #342


sylvia-derive/src/input.rs
```diff
@@ -173,6 +173,8 @@ pub struct ImplInput<'a> {
 
 impl<'a> ImplInput<'a> {
     pub fn new(item: &'a ItemImpl) -> Self {
+        assert_new_method(item);
```
kulikthebird, line 176:
```suggestion
        assert_new_method_defined_for_contract(item);
```

sylvia-derive/src/input.rs
```diff
@@ -289,3 +291,26 @@ impl<'a> ImplInput<'a> {
         ContractMtHelpers::new(item, generic_params, custom, override_entry_points.clone()).emit()
     }
 }
+
+fn assert_new_method(item: &ItemImpl) {
```
kulikthebird, line 295: It seems like this method should be part of the parser module

kulikthebird (changes requested):
Useful error. One comment, lgtm

kulikthebird (approved):

PR #347, opened 2024-04-10
build: Add codecov token for upload

jawoznia:


kulikthebird (approved):
Thanks for fixing that!

PR #353, opened 2024-04-17
chore: Unify MT proxy methods emit

jawoznia:
Cleanup in `Multitest` generation. After removing the support for the `contract` macro on `impl Interface` we can merge some methods as they do not differ this much between `interface` and `contract`.


kulikthebird (approved):
Decreasing the complexity of `message` module is definitely a good direction.

PR #358, opened 2024-04-18
feat: No need to pass types to non-generic interfaces

kulikthebird:


kulikthebird:
There's a new proposal here: https://github.com/CosmWasm/sylvia/pull/361

Issue #374, opened 2024-05-08
Allow sylvia crates to use cw-orchestrator as well


ethanfrey:
Sylvia is a powerful tool to make writing and composing contracts faster and safer, and also includes some great test helpers for multi-test based testing.

[cw-orchestrator](https://orchestrator.abstract.money/) is a powerful tool to allow rust based testing and deployment. It starts where Sylvia stops (they both cover multitest), and then allows running real tests on local-net or osmosis test tube, as well as deploying to testnet or mainnet... all with the same, well-typed, Rust code.

As I architect the progress of mesh security, I would love to have the ability to sue cw-orch as well in this Sylvia codebase. Unfortunatley, this is currently impossible as cw-orchestrator [requires a few derive macros](https://orchestrator.abstract.money/contracts/entry-points.html) on the ExecuteMsg and QueryMsg, which are auto-generated by sylvia and thus inaccessible.

Furthermore, I note that as of 0.22, `cw-orch` is on CosmWasm 1.x, so any integration should target the 0.10.x (or 0.11.x?) line of Sylvia.

Pinging @jawoznia and @hashedone for visibility on the Confio side

Pinging @kayanski and @adairrr for visibility on the Abstarct side.

I note this is my idea and I have gotten some support from both teams to move it forward, but I am the requestor

adairrr:
Confirming visibility! This would be a great addition to the CosmWasmers' toolkit, as it also unifies the testing suites between traditional CosmWasm contracts and Sylvia-based contracts.

hashedone:
I've looked at the PR, and I don't like the idea of bounding the dependencies of Sylvia and cw-orchestrator. So many things can go wrong on this level that I believe it only looks simple. You just mentioned that we would need to target the Sylvia 0.10.0 line with that (that is not meant to have more features; it is only for bugfix backporting), but in fact, examples in Orchestrator still use cw 1.4, while Sylvia 0.10.0 targets 1.5. I have no idea if that would or would not cause issues. And this would have to be reensured on every sylvia release, orch release, and cw release - I find this unmaintainable.
Instead, we prefer to go more generic with the simpler design that simply forwards attributes to generated entities: #379. I mentioned it when we discussed before. This solution is both simpler and better at firing the Sylvia plans.
Also, after thinking of this I am really not a fan of adding this feature to 0.x line, and there are two reasons three reasons for that:
* It adds maintenance complexity - there are more features in the soon-to-be-deprecated version of a crate
[...]

PR #375, opened 2024-05-08
Demo: add cw-orch support

ethanfrey:
This is a demo PR to show the possibility of implementing #374 

It has some short-comings, but in the current state, it allows crates to optionally enable cw_orch derives in sylvia and use it in their crate. While leaving crates that don't opt-in with the same code-gen as before (meaning it is potentially non-API breaking).

In this case, @kayanski modified the cw20-base sylvia example to play well with the cw-orch framework and we run a multi-test on the cw-orch generated code to show it works.

I will highlight some limitations of the current PR that need to be fixed before such could be merged, but this should provide a concrete example to move any potential cw-orch and sylvia collaboration forward

### DIFF
--- a/examples/contracts/cw20-base/src/orch.rs
+++ b/examples/contracts/cw20-base/src/orch.rs
@@ -133,9 +133,19 @@ mod tests {
 
         // transfer some tokens to self
         contract
-            .transfer(Uint128::new(1_000_000), random.to_string())
+            .transfer(
+                Uint128::new(1_000_000),
+                random.to_string(),
+                &[coin(300_000, denom)],
+            )
             .unwrap();
         let bal = contract.balance(random.to_string()).unwrap();
         assert_eq!(bal.balance.u128(), 1_000_000);
+
+        let native = chain
+            .balance(&chain.sender(), Some(denom.to_string()))
+            .unwrap();
+        assert_eq!(native.len(), 1);
+        assert_eq!(native[0].amount.u128(), 700_000);
     }
 }
--- a/examples/contracts/cw20-base/tests/cw-orch.rs
+++ b/examples/contracts/cw20-base/tests/cw-orch.rs
@@ -41,20 +41,13 @@ fn mock_interact() -> cw_orch::anyhow::Result<()> {
     // TODO: how do i get a different address for testing???
     // Here I just send to self
     let friend = MockBech32::new("mock");
-    contract.transfer(10_000u128.into(), friend.sender().to_string())?;
+    contract.transfer(10_000u128.into(), friend.sender().to_string(), &[])?;
     assert_eq!(friend.sender().to_string(), mock.sender().to_string());
     let balance = contract.balance(mock.sender().to_string())?;
     assert_eq!(balance.balance.u128(), 150_000u128);
 
-    // This is what fails... let's see how to make that happen
-    contract.mint(150_000u128.into(), mock.sender().to_string())?;
-    // contract.execute(
-    //     &ContractExecMsg::Minting(Cw20MintingExecMsg::Mint {
-    //         recipient: mock.sender().to_string(),
-    //         amount: 150_000u128.into(),
-    //     }),
-    //     None,
-    // )?;
+    // Call mint
+    contract.mint(150_000u128.into(), mock.sender().to_string(), &[])?;
     let balance = contract.balance(mock.sender().to_string())?;
     assert_eq!(balance.balance.u128(), 300_000u128);
 
--- a/sylvia-derive/src/message.rs
+++ b/sylvia-derive/src/message.rs
@@ -381,7 +381,6 @@ impl<'a> ContractEnumMessage<'a> {
             false => quote! {},
         };
 
-
         quote! {
             #[allow(clippy::derive_partial_eq_without_eq)]
             #[derive(#sylvia ::serde::Serialize, #sylvia ::serde::Deserialize, Clone, Debug, PartialEq, #sylvia ::schemars::JsonSchema, #derive_query )]
@@ -483,6 +482,7 @@ impl<'a> MsgVariant<'a> {
         let fields = fields.iter().map(MsgField::emit);
         let returns_attribute = match msg_type {
             MsgType::Query => quote! { #[returns(#stripped_return_type)] },
+            MsgType::Exec => quote! { #[payable] }, // for cw-orch, ideally dynamically derived
             _ => quote! {},
         };
 
@@ -1277,7 +1277,7 @@ impl<'a> GlueMessage<'a> {
         let ep_name = msg_ty.emit_ep_name();
         let messages_fn_name = Ident::new(&format!("{}_messages", ep_name), contract.span());
         let contract_variant = quote! { #contract_name ( #enum_name #bracketed_used_generics ) };
-        
+
         let mut messages_call = interfaces.emit_messages_call(msg_ty);
         let prefixed_used_generics = if !used_generics.is_empty() {
             quote! { :: #bracketed_used_generics }


Cargo.toml
```diff
@@ -8,6 +8,7 @@ version = "0.10.1"
 
 [workspace.dependencies]
 sylvia-derive = { version = "0.10.1", path = "sylvia-derive" }
+cw-orch = { git = "https://github.com/abstractsdk/cw-orchestrator", branch = "update/relax-bound-on-derive" }
```
ethanfrey, line 11: Obviously, this would need to be merged and tagged before we can accept a PR, but that should be doable.

ethanfrey: The bigger question is why we have to tie Sylvia (even as an optional dependency) to one version of cw-orch. (eg 0.10x sylvia uses cw-orch 0.22, 0.11.x sylvia uses cw-orch 0.23). Maybe someone with more macro magic than I can see how to solve that.

examples/contracts/cw20-base/src/orch.rs
```diff
@@ -0,0 +1,89 @@
+// This contains all code we needed to manually add to make it work well with cw-orch
+// In the future, hopefully some of this can me auto-generated. But let's get it to work now.
+use crate::contract::entry_points::{execute, instantiate, query};
+use crate::contract::sv::{ContractExecMsg, ContractQueryMsg, ContractSudoMsg, InstantiateMsg};
+use cw_orch::prelude::*;
+
+// Maybe uploadable can be autogenerated?
+// But this is fine to include in the client code
+
+#[cw_orch::interface(InstantiateMsg, ContractExecMsg, ContractQueryMsg, Empty)]
```
ethanfrey, line 10: We can try to auto-generate this, or just leave this done by hand in the contracts that enable the "orch" feature

examples/contracts/cw20-base/src/orch.rs
```diff
@@ -0,0 +1,89 @@
...
+// The following is done in Sylvia, ideally the other (interface) variants could be added there too:
+// impl From<ExecMsg> for ContractExecMsg
+
+
+// Since we know the regular structure here, this may be easy enough to
+// generate inside the `sv::messages` macro rather than the `contract` macro, eg.
+// #[sv::messages(cw20_marketing as Marketing)]
+// #[sv::messages(cw20_minting as Minting)]
+
+impl From<cw20_allowances::sv::Cw20AllowancesExecMsg> for ContractExecMsg {
```
ethanfrey, line 37: Ideally Sylvia could generate these for all the interfaces that are used as `ContractExecMsg` variants. (Help wanted... the commit by @Kayanski shows the limit of trait associated types.)

examples/contracts/cw20-base/tests/cw-orch.rs
```diff
@@ -0,0 +1,62 @@
...
+                    amount: 150_000u128.into(),
+                }],
+                marketing: None,
+            },
+        },
+        None,
+        None,
+    )?;
+
+    let balance = contract.balance(mock.sender().to_string())?;
```
ethanfrey, line 38: Calling the contract with nice syntax. 
That works on multi-test, testnet, etc.

:+1:

examples/contracts/cw20-base/tests/cw-orch.rs
```diff
@@ -0,0 +1,62 @@
...
+
+    // TODO: how do i get a different address for testing???
+    // Here I just send to self
+    let friend = MockBech32::new("mock");
+    contract.transfer(10_000u128.into(), friend.sender().to_string())?;
+    assert_eq!(friend.sender().to_string(), mock.sender().to_string());
+    let balance = contract.balance(mock.sender().to_string())?;
+    assert_eq!(balance.balance.u128(), 150_000u128);
+
+    // This is what fails... let's see how to make that happen
```
ethanfrey, line 49: Old comment, I fixed this... that is why the hand-rolled "From" statements are needed. Can we agree the new form is cleaner?

examples/interfaces/cw20-allowances/Cargo.toml
```diff
@@ -16,6 +16,7 @@ cosmwasm-std = { workspace = true, features = ["staking"] }
 cosmwasm-schema = { workspace = true }
 serde = { workspace = true }
 sylvia = { path = "../../../sylvia" }
+cw-orch = { workspace = true }
```
ethanfrey, line 19: This is a bit of a trick.

All the interfaces (or actually any sylvia crate) that will be imported by a `cw-orch`-enabled sylvia contract will need to have this dependency. No code changes, just the dependency. The macro exansion enabled by the `orch` feature ends up referencing `::cw_orch::...`. Maybe there is a way around it?

Only seems a problem if you import interface crates that are not in your repo

sylvia-derive/src/message.rs
```diff
@@ -354,8 +354,19 @@ impl<'a> ContractEnumMessage<'a> {
         let ctx_type = msg_ty.emit_ctx_type(&custom.query_or_default());
         let ret_type = msg_ty.emit_result_type(&custom.msg_or_default(), &error.error);
 
+        #[cfg(feature = "orch")]
```
ethanfrey, line 357: Change #1: add derives to the messages

sylvia-derive/src/message.rs
```diff
@@ -1382,7 +1395,23 @@ impl<'a> GlueMessage<'a> {
                     Err(D::Error::custom(err_msg))
                 }
             }
+        };
+
+        // Append this derive for cw-orch contracts
+        // (Optionally you could always do this?)
+        #[cfg(feature = "orch")]
```
ethanfrey, line 1402: Change #2: Add the From statement for the contract specific message.

(Note: we want to extend this for all the messages from all the interface variants, but that proved hard)

sylvia-derive/src/message_type.rs
```diff
@@ -161,4 +162,20 @@ impl MsgType {
             },
         }
     }
+
+    #[cfg(feature = "orch")]
+    pub fn emit_derive_call(&self) -> TokenStream {
```
ethanfrey, line 167: Change #3: add those derive bounds on other messages

sylvia/src/lib.rs
```diff
@@ -27,3 +27,5 @@ pub use {
     cosmwasm_schema as cw_schema, cosmwasm_std as cw_std, schemars, serde,
     serde_cw_value as serde_value, serde_json_wasm as serde_json,
 };
+#[cfg(feature = "orch")]
+pub use cw_orch;
```
ethanfrey, line 31: Conditionally export cw-orch here so that those lines like `#sylvia:: cw_orch::QueryFns` compile

ethanfrey (commented):
Left some comments on the state of this POC.

How it works, and what further work would be needed.
I'd love responses on these points

sylvia/src/lib.rs
```diff
@@ -27,3 +27,5 @@ pub use {
     cosmwasm_schema as cw_schema, cosmwasm_std as cw_std, schemars, serde,
     serde_cw_value as serde_value, serde_json_wasm as serde_json,
 };
+#[cfg(feature = "orch")]
```
jawoznia, line 30: This will make it visible in docs as reexported if `cw_orch` is enabled.

```suggestion
#[cfg_attr(docsrs, doc(cfg(feature = "cw_orch")))]
#[cfg(feature = "orch")]
```

sylvia-derive/src/message_type.rs
```diff
@@ -161,4 +162,20 @@ impl MsgType {
...
+                #[derive(#sylvia ::serde::Serialize, #sylvia ::serde::Deserialize, Clone, Debug, PartialEq, #sylvia ::schemars::JsonSchema, #sylvia:: cw_schema::QueryResponses, #sylvia:: cw_orch::QueryFns)]
+            },
+            MsgType::Exec => quote! {
+                #[derive(#sylvia ::serde::Serialize, #sylvia ::serde::Deserialize, Clone, Debug, PartialEq, #sylvia ::schemars::JsonSchema, #sylvia:: cw_orch::ExecuteFns)]
+            },
+            _ => quote! {
+                #[derive(#sylvia ::serde::Serialize, #sylvia ::serde::Deserialize, Clone, Debug, PartialEq, #sylvia ::schemars::JsonSchema)]
+            },
+        }
+    }
```
jawoznia, line 180: No need for such split.

Add method in `MsgType` to return `#sylvia:: cw_orch::QueryFns` or `#sylvia:: cw_orch::ExecuteFns` for `MsgType::Query/Exec` respectively if flag is enabled.

sylvia-derive/src/message.rs
```diff
@@ -354,8 +354,19 @@ impl<'a> ContractEnumMessage<'a> {
...
+                quote! { #sylvia ::cw_schema::QueryResponses, #sylvia:: cw_orch::QueryFns }
+            }
+            MsgType::Exec => quote! { #sylvia:: cw_orch::ExecuteFns },
+            _ => quote! {},
+        };
+        #[cfg(not(feature = "orch"))]
+        let derive_query = match msg_ty {
+            MsgType::Query => {
+                quote! { #sylvia ::cw_schema::QueryResponses }
+            }
```
jawoznia, line 369: Same as in case of `emit_derive`

hashedone (commented):
I understand the need and generalized it in #379 issue that should solve the problem.

The big issue I have here is binding the Sylvia implementation to particular external library. I am a fan of providing the API to integrate them, but I really don't like the idea of API saying "you can use this feature, but then you need to add additional dependency and it has to be in particular version". This adds a lot of potential maintenance work - because, at some point, some bug might occur, saying, "I use Sylvia x.y.z with orchestrator a.b.c, and it doesn't work.".
Providing a way to simply forward attributes to messages doesn't create artificial dependency corelations.

ethanfrey:
> Providing a way to simply forward attributes to messages doesn't create artificial dependency corelations.

I support this approach 100% and was struggling to remove the import, which I couldn't (and signalled as an issue).

I have no idea how one can extend remote macros. But if there is a way, and it can cover the code points here, it would be great. And this serves as an example of what places would need to be extended.

Also if #376 is implement, the only points left to extend are just adding some derive macros on top of the Execute, Query, Sudo messages, which should be a reasonable target. And a lot more clear than "allow others to extend".

Very happy to close this PR and use another approach if it can acheive this outcome

CyberHoward:
> This adds a lot of potential maintenance work - because, at some point, some bug might occur, saying, "I use Sylvia x.y.z with orchestrator a.b.c, and it doesn't work.". Providing a way to simply forward attributes to messages doesn't create artificial dependency correlations.

This is a great point that we identified a while ago. To solve this we've stabilized the core traits and types used in cw-orch. See the [cw-orch-core](https://crates.io/crates/cw-orch-core) crate. 

This change essentially means that an interface with cw-orch v0.20.x is compatible with a version of v0.21.x and above until the next breaking change to core (which should be rare). This means that not all crates have to be on the same version of cw-orch.

Lmk if you think we're missing something with this or if there's a better way for us to make this clear to the crate consumer!

kulikthebird:
Hi Ethan,

Would you agree to close this demo PR? We already have two issues that specify the problem demonstrated in this PR:
 * https://github.com/CosmWasm/sylvia/issues/374
 * https://github.com/CosmWasm/sylvia/issues/376
 
The first of these is already addressed by https://github.com/CosmWasm/sylvia/pull/388.

jawoznia:
> Hi Ethan,
> 
> Would you agree to close this demo PR? We already have two issues that specify the problem demonstrated in this PR:
> 
>     * [Allow sylvia crates to use cw-orchestrator as well #374](https://github.com/CosmWasm/sylvia/issues/374)
> 
>     * [Add impl From<VARIANTS> for Contract{Exec,Query,Sudo}Msg #376](https://github.com/CosmWasm/sylvia/issues/376)
> 
> 
> The first of these is already addressed by #388.

@ethanfrey, did you check if the attributes forwarding satisfy the cw-orch support?

jawoznia:
Closing as this PR should be resolved by the #374 and #376.

If there will be missing support for attributes forwarding please create an issue.

PR #367, opened 2024-05-07
fix: Allow storage of `Remote<dyn Interface>`

jawoznia:

PR #378, opened 2024-05-09
test: Store Remote contract and communicate with it

jawoznia:


sylvia/tests/remote.rs
```diff
@@ -1,62 +1,379 @@
...
 
 // Making sure `Remote` can be stored in `#[cw_serde]` types
 #[cw_serde]
 #[allow(dead_code)]
-struct CustomStorage<'a, Contract> {
-    remote: sylvia::types::Remote<'a, Contract>,
-    interface_remote:
-        sylvia::types::Remote<'static, &'a dyn some_interface::SomeInterface<Error = StdError>>,
+pub struct ContractStorage<'a, Contract> {
+    remote: Remote<'a, Contract>,
```
hashedone, line 165: Having an example with the contract is good, but we still need working remotes for interfaces. It should not require reference.

hashedone (commented):
I dont really understand this change - is it only the tests upgrade?

sylvia/tests/remote.rs
jawoznia: Fixed in #367. Will rebase after it is merged.

kulikthebird (approved):

PR #382, opened 2024-05-22
feat: No ref needed for 'dyn Interface' type in Remote and BoundQuerier

kulikthebird:


jawoznia (approved):
Port also the #378
LGTM

PR #393, opened 2024-07-10
chore: Update documentation and refactoring

kulikthebird:


README.md
```diff
@@ -292,6 +298,10 @@ impl MyContract {
  * `sv::custom` allows to define CustomMsg and CustomQuery for the contract. By default generated code
     will return `Response<Empty>` and will use `Deps<Empty>` and `DepsMut<Empty>`.
 
+ * `sv::msg_attr` forwards any attribute to the message's type.
+
+ * `sv::attr` forwards any attribute to the enum's field of the given variant.
```
jawoznia, line 303: Forwarding for enum variant fields is done without any attribute

```suggestion
 * `sv::attr` forwards any attribute to the enum's variant.
```

README.md
```diff
@@ -378,6 +388,24 @@ pub fn evaluate_member(&self, ctx: ExecCtx, ...) -> StdResult<Response> {
 }
 ```
 
+
+## Executor message builder
+
+Execute messages of other contracts can be built with `Remote` as well by
+calling `executor` method. It returns a message builder that implements
+auto-generated `Executor` traits of all sylvia contracts.
```
jawoznia, line 396: Since the `ExecutorBuilder` is generic over a contract this type implements only the traits of this contract and all the interfaces it implements.

I might over expand on this for README, but feel free to consider some parts.

```suggestion
Sylvia defines the [`ExecutorBuilder`]() type, which can be accessed through [`Remote::executor`]().
It's generic over the contract type and exposes execute methods from the contract and every interface implemented on it through an auto-generated `Executor` traits.
Execute messages of other contracts can be built with `Remote` as well by
calling `executor` method. It returns a message builder that implements
auto-generated `Executor` traits of all Sylvia contracts.

Methods defined in the `Executor` traits constructs an execute message, which variant corresponds to the method name.
The message is then wrapped in the `WasmMsg`, and returned once [`ExecutorBuilder::build()`]() method is called.
```

sylvia-derive/src/lib.rs
```diff
@@ -227,6 +227,20 @@ pub(crate) fn crate_module() -> Path {
 /// # fn main() {}
 /// ```
 ///
+/// ### `sv::msg_attr(msg_type, {...})`
+///
+/// This attribute can be used for the whole `trait Interface {}` block and
+/// for the following message types: `exec`, `query`, `sudo` and `reply`.
```
jawoznia, line 233:
```suggestion
/// for the following message types: `exec`, `query` and `sudo`.
```

jawoznia (commented):
Some suggestions to consider

jawoznia (approved):

Issue #399, opened 2024-07-18
Make cw-* examples generic


hashedone:
The CW-* examples are straightforwardly copied from the cw-plus repository, but they have a flaw - they all use the `Empty` custom msg (and query), which is not the best choice.

The idea is that contracts/interfaces that use `Empty` are type-wise meant to be uploaded into the chains that do not use `Empty` messages. They would work on other chains, as they do not send custom messages, so they do not cause invalid execution. Still, the same is true for any contract that produces an invalid custom message as long as it produces only variants supported by a particular chain.

The issue with this approach is testing - the MT environment is strongly typed, including the custom exec/query. It is its strength, not a weakness, and the best is to use it to enforce good programming practices instead of workarounds. The contract that is meant to be used on chains with not determined customs should be a contract that has generic customs - and that should be the *default* way of implementing contracts (and interfaces as well - those should, however, use associated types for generalization).

However, the examples we use do not show this practice - instead, they use `Empty,` disregarding the fact that one might want to use it as part of a wider testing environment with non-empty custom messages. They should all be changed to use the recommended way of building chains-agnostic contracts.

## Interfaces
- [ ] cw1
- [ ] cw4
- [ ] cw20-allowances
- [ ] cw20-marketing
- [ ] cw20-minting
- [ ] whitelist

## Contracts
- [ ] cw1-subkeys
- [ ] cw1-whitelist
- [ ] cw20-base

PR #402, opened 2024-07-19
Generic interfaces on examples

abefernan:
Part of #399.

This PR deals with the "Interfaces" part of the issue above.

It also makes the affected contracts and tests use `cosmwasm_std::Empty` for their different implementations of the now generic interfaces.

### DIFF
--- a/examples/contracts/custom/src/cw1.rs
+++ b/examples/contracts/custom/src/cw1.rs
@@ -1,4 +1,4 @@
-use cosmwasm_std::{CosmosMsg, Response, StdError, StdResult};
+use cosmwasm_std::{CosmosMsg, Empty, Response, StdError, StdResult};
 use cw1::{CanExecuteResp, Cw1};
 use sylvia::types::{ExecCtx, QueryCtx};
 
@@ -6,6 +6,9 @@ use crate::contract::CustomContract;
 
 impl Cw1 for CustomContract {
     type Error = StdError;
+    type ExecC = Empty;
+    type QueryC = Empty;
+    type CosmosCustomMsg = Empty;
 
     fn execute(&self, _ctx: ExecCtx, _msgs: Vec<CosmosMsg>) -> StdResult<Response> {
         Ok(Response::new())
--- a/examples/contracts/cw1-subkeys/src/cw1.rs
+++ b/examples/contracts/cw1-subkeys/src/cw1.rs
@@ -1,4 +1,4 @@
-use cosmwasm_std::{ensure, Addr, Response, StdResult};
+use cosmwasm_std::{ensure, Addr, Empty, Response, StdResult};
 use cw1::{CanExecuteResp, Cw1};
 use sylvia::types::{ExecCtx, QueryCtx};
 
@@ -7,6 +7,9 @@ use crate::error::ContractError;
 
 impl Cw1 for Cw1SubkeysContract {
     type Error = ContractError;
+    type ExecC = Empty;
+    type QueryC = Empty;
+    type CosmosCustomMsg = Empty;
 
     fn execute(
         &self,
--- a/examples/interfaces/cw1/src/lib.rs
+++ b/examples/interfaces/cw1/src/lib.rs
@@ -1,6 +1,6 @@
 use cosmwasm_std::{CosmosMsg, Response, StdError, StdResult};
 use serde::{Deserialize, Serialize};
-use sylvia::types::{ExecCtx, QueryCtx};
+use sylvia::types::{CustomMsg, CustomQuery, ExecCtx, QueryCtx};
 use sylvia::{interface, schemars};
 
 #[derive(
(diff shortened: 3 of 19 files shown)


examples/contracts/cw1-subkeys/src/cw1.rs
```diff
@@ -7,6 +7,9 @@ use crate::error::ContractError;
 
 impl Cw1 for Cw1SubkeysContract<'_> {
```
hashedone, line 8: Just to confirm - the contract should be generic over exec and query, but I assume there would be separate PR for that, right?

examples/contracts/custom/src/cw1.rs
```diff
@@ -1,11 +1,14 @@
...
 use cw1::{CanExecuteResp, Cw1};
 use sylvia::types::{ExecCtx, QueryCtx};
 
 use crate::contract::CustomContract;
 
 impl Cw1 for CustomContract {
     type Error = StdError;
+    type ExecC = Empty;
+    type QueryC = Empty;
+    type CosmosCustomMsg = Empty;
```
hashedone, line 11: What is this one about? @jawoznia do we need this for anything? I see it in more places so maybe I miss something?

hashedone (approved):
In general looks ok, but I see couple `CosmosCustomMsg` and I don't know what are they about. Do we need those?

examples/contracts/cw1-subkeys/src/cw1.rs
abefernan: Yes, this PR only deals with interfaces

examples/contracts/custom/src/cw1.rs
jawoznia: `ExecC` and `QueryC` are special associated types by which you can define the `CustomMsg` and `CustomQuery` types.

The `CosmosCustomMsg` is not needed here judging by the name.

examples/interfaces/cw1/src/lib.rs
```diff
@@ -11,35 +11,43 @@ pub struct CanExecuteResp {
...
 
     /// Execute requests the contract to re-dispatch all these messages with the
     /// contract's address as sender. Every implementation has it's own logic to
     /// determine in
     #[sv::msg(exec)]
-    fn execute(&self, ctx: ExecCtx, msgs: Vec<CosmosMsg>) -> Result<Response, Self::Error>;
+    fn execute(
+        &self,
+        ctx: ExecCtx<Self::QueryC>,
+        msgs: Vec<CosmosMsg<Self::CosmosCustomMsg>>,
```
jawoznia, line 27:
```suggestion
        msgs: Vec<CosmosMsg<Self::ExecC>>,
```

examples/interfaces/cw1/src/lib.rs
```diff
@@ -11,35 +11,43 @@ pub struct CanExecuteResp {
 }
 
 #[interface]
-#[sv::custom(msg=cosmwasm_std::Empty, query=cosmwasm_std::Empty)]
 pub trait Cw1 {
     type Error: From<StdError>;
+    type ExecC: CustomMsg;
+    type QueryC: CustomQuery;
+    type CosmosCustomMsg: CustomMsg;
```
jawoznia, line 18:
```suggestion
```

jawoznia (commented):
`ExecC` is a type meant to be used for defining the custom message in the interface.

examples/contracts/custom/src/cw1.rs
abefernan: As discussed in Slack, it seems needed for now since it wouldn't compile.

abefernan:
> In general looks ok, but I see couple `CosmosCustomMsg` and I don't know what are they about. Do we need those?

It seems like it's needed for now or else it won't compile. @jawoznia and @kulikthebird think it might be a bug in Sylvia

examples/interfaces/cw1/src/lib.rs
jawoznia, line 27: Didn't we want to make it

```suggestion
        msgs: Vec<CosmosMsg<Self::ExecC>>,
```

examples/interfaces/cw1/src/lib.rs
```diff
@@ -11,35 +11,43 @@ pub struct CanExecuteResp {
...
     /// If CanExecute returns true then a call to `Execute` with the same message,
     /// from the given sender, before any further state changes, should also succeed.
     #[sv::msg(query)]
     fn can_execute(
         &self,
-        ctx: QueryCtx,
+        ctx: QueryCtx<Self::QueryC>,
         sender: String,
-        msg: CosmosMsg,
+        msg: CosmosMsg<Self::CosmosCustomMsg>,
```
jawoznia, line 38: same `ExecC`

examples/interfaces/cw4/src/lib.rs
```diff
@@ -1,44 +1,81 @@
...
-#[sv::custom(msg=cosmwasm_std::Empty, query=cosmwasm_std::Empty)]
 pub trait Cw4 {
     type Error: From<StdError>;
+    type ExecC: CustomMsg;
+    type QueryC: CustomQuery;
+    type MemberCustomMsg: CustomMsg;
+    type ListMembersCustomMsg: CustomMsg;
+    type TotalWeightCustomMsg: CustomMsg;
+    type AdminCustomMsg: CustomMsg;
+    type HooksCustomMsg: CustomMsg;
```
jawoznia, line 14: Remove.

The queries most likely should not return the `Response` and instead some custom response type

Please add those messages
https://github.com/CosmWasm/cw-plus/blob/main/packages/cw4/src/query.rs

jawoznia (changes requested):

examples/interfaces/cw1/src/lib.rs
abefernan: If I change this line I get `associated type ExecC not found for Self`.

And also the `#[interface]` directive errors with:

```
type alias takes 1 generic argument but 0 generic arguments were supplied
expected 1 generic argumentrustc
lib.rs(13, 1): type alias defined here, with 1 generic parameter: `ExecC`
lib.rs(13, 2): add missing generic argument: `ExecC`
```

Do I need to use the sv::custom directive?

jawoznia: Ok, I thought the [fix](https://github.com/CosmWasm/sylvia/pull/406) landed in this PR.

Leave it as it is and remove `CosmosCustomMsg` in following PR then.

jawoznia (approved):
LGTM

PR #403, opened 2024-07-22
feat: Deduce generic custom types in entry_points macro

jawoznia:


hashedone (approved):
LGTM, but I see coverage dropped. Could you add some test cases to make s happy?

PR #406, opened 2024-07-29
feat: Allow usage of `ExecC` and `QueryC` in other params then `Response` and `Ctx` types

jawoznia:


hashedone (approved):
Good important change

PR #409, opened 2024-08-01
fix: Pass attribute to struct fields

jawoznia:


sylvia-derive/src/message.rs
```diff
@@ -791,6 +790,21 @@ impl<'a> MsgField<'a> {
         }
     }
 
+    /// Emits struct field
+    pub fn emit_pub(&self) -> TokenStream {
```
kulikthebird, line 794: Let's give it some more explanatory name

sylvia/tests/messages_generation.rs
```diff
@@ -201,7 +205,9 @@ fn contract_messages_constructible() {
     let _ = contract::sv::SudoMsg::ArgumentedSudo {
         _user: Addr::unchecked("owner"),
     };
-    let _ = contract::sv::InstantiateMsg {};
+    let _ = contract::sv::InstantiateMsg {
```
kulikthebird, line 208: How does this test check that sylvia forwarded the attribute to the msg field?

sylvia/tests/messages_generation.rs
```diff
@@ -304,7 +312,13 @@ fn attributes_forwarding_for_message_variants() {
         .to_string(),
         "ArgumentedSudo"
     );
-    assert_eq!(contract::sv::InstantiateMsg {}.to_string(), "Instantiate");
+    assert_eq!(
```
kulikthebird, line 315: Same as above

kulikthebird (changes requested):

sylvia-derive/src/message.rs
jawoznia: Tbh I don't see a better name here. It differs from the `emit` method only in adding the `pub` before the field.

We can rethink the naming of those methods, but this would affect every emitter in the project so let's do that outside of this PR.

kulikthebird (approved):
Nice!

PR #411, opened 2024-08-05
chore: Setup new directory structure

jawoznia:


kulikthebird (approved):
That's a good change, definitely a step in the right direction. I like the separation between the three types of macros that Sylvia defines. Let's consider this a first step and look to add further distinctions between the new modules, especially the parsers.

PR #412, opened 2024-08-06
chore: Refactor struct message generation

jawoznia:


sylvia-derive/src/contract/communication/struct_msg.rs
```diff
@@ -32,138 +26,109 @@ impl<'a> StructMessage<'a> {
...
             #[derive(#sylvia ::serde::Serialize, #sylvia ::serde::Deserialize, Clone, Debug, PartialEq, #sylvia ::schemars::JsonSchema)]
             #( #[ #msg_attrs_to_forward ] )*
             #[serde(rename_all="snake_case")]
-            pub struct #name #generics {
+            pub struct #name #bracketed_used_generics {
                 #(#fields,)*
             }
 
-            impl #generics #name #generics #where_clause {
+            impl #bracketed_used_generics #name #bracketed_used_generics #where_clause {
```
kulikthebird, line 126: The set of generics have been changed here - it'd be beneficial to see that change somewhere in the test cases

kulikthebird (changes requested):
Looks good! Only one comment related to the change of the generated message structure. Also it seems like this change can be breaking

sylvia-derive/src/contract/communication/struct_msg.rs
jawoznia: It didn't. The previous naming was incorrect.

```rust
generics: used_generics,
```

kulikthebird (approved):

PR #413, opened 2024-08-07
chore!: Remove `InterfaceApi` in favor of `InterfaceMessagesApi`

kulikthebird:


jawoznia (changes requested):
Nice, but please mark this change as breaking in commit message

sylvia-derive/src/interface/communication/api.rs
```diff
@@ -58,45 +49,27 @@ impl<'a> Api<'a> {
...
-                type Querier<'querier> = #sylvia ::types::BoundQuerier<'querier, #custom_query, Contract >;
-            }
-
-
-            pub struct Api < #(#generics,)* > {
-                #phantom
             }
 
-            impl < #(#generics,)* > #sylvia ::types::InterfaceApi for Api < #(#generics,)* > #where_clause {
+            impl<'a, Error, #(#generics),*> InterfaceMessagesApi for dyn #interface_name < Error = Error, #(#generics = #generics,)* > + 'a {
```
jawoznia, line 69: It should not be an issue here, but please keep the convention of naming generated lifetimes and make it more unique with `sv_` prefix

jawoznia (approved):
One comment. Otherwise LGTM

sylvia-derive/src/interface/communication/api.rs
kulikthebird: Done

PR #419, opened 2024-08-23
feat: Make user side cosmwasm-std dep obsolete

jawoznia:
Provide path to sylvia exposed cw_std dependency in entry_point macro


kulikthebird (approved):
Looks good!

PR #421, opened 2024-08-23
feat: Make cosmwasm_schema deps obsolete

jawoznia:
This makes `cosmwasm_schema` dependency optional on user side, however due to how `cosmwasm_schema` and `schemars` is constructed it requires user to specify path to add some more info to macros like so:

```rust
#[cw_serde(crate = "sylvia::cw_schema")]
pub struct MyType {}
```

```rust
write_api! {
    crate_name: sylvia::cw_schema,

    instantiate: InstantiateMsg,
    execute: ContractExecMsg<Empty, Empty>,
    query: ContractQueryMsg<Empty, Empty>,
}
```

Users however can still add `cosmwasm_schema` to their deps and this change allows them to skip this dependency if they want to.

### DIFF
--- a/examples/Cargo.lock
+++ b/examples/Cargo.lock
@@ -406,7 +406,6 @@ name = "custom"
 version = "0.5.0"
 dependencies = [
  "anyhow",
- "cosmwasm-schema",
  "cw-multi-test",
  "cw-storage-plus",
  "cw1",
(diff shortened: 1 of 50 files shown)


kulikthebird (approved):
:+1:

PR #422, opened 2024-08-26
feat: Make serde dep user side obsolete

jawoznia:
Removed `serde` and `cw_multi_test` dependency from examples.
Fixed features in examples as some already non existent features were used in `cfg` checks. Also `mt` feature now enables `sylvia/mt` instead of `cw_multi_test`.

### DIFF
--- a/examples/contracts/custom/Cargo.toml
+++ b/examples/contracts/custom/Cargo.toml
@@ -19,12 +19,10 @@ mt = ["library", "anyhow", "cw-multi-test"]
 [dependencies]
 cw1 = { path = "../../interfaces/cw1" }
 cw-storage-plus = { workspace = true }
-serde = { workspace = true }
 sylvia = { path = "../../../sylvia" }
 cw-multi-test = { workspace = true, optional = true }
 anyhow = { workspace = true, optional = true }
 
 [dev-dependencies]
 anyhow = { workspace = true }
-cw-multi-test = { workspace = true }
 sylvia = { path = "../../../sylvia", features = ["mt"] }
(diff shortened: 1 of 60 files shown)


examples/contracts/entry-points-overriding/Cargo.toml
```diff
@@ -13,17 +13,15 @@ crate-type = ["cdylib", "rlib"]
 
 [features]
 library = []
-mt = ["library", "cw-multi-test", "anyhow"]
+mt = ["library", "sylvia/mt", "anyhow"]
 
 [dependencies]
 anyhow = { workspace = true, optional = true }
 cw-multi-test = { workspace = true, optional = true }
```
kulikthebird, line 20: Do we still need this?

examples/contracts/custom/Cargo.toml
```diff
@@ -19,12 +19,10 @@ mt = ["library", "anyhow", "cw-multi-test"]
 [dependencies]
 cw1 = { path = "../../interfaces/cw1" }
 cw-storage-plus = { workspace = true }
-serde = { workspace = true }
 sylvia = { path = "../../../sylvia" }
 cw-multi-test = { workspace = true, optional = true }
```
kulikthebird, line 23: And this

kulikthebird (approved):
Nice! Two comments to fix and we're good to go

examples/contracts/entry-points-overriding/Cargo.toml
jawoznia: The `anyhow` and `multitest` deps are removed in the following PR. I missed them here

examples/contracts/custom/Cargo.toml
jawoznia: same

PR #438, opened 2024-09-27
feat: Add `UpdateAdmin` and `ClearAdmin` constructors in the `Remote`

jawoznia:


sylvia/tests/remote.rs
```diff
@@ -331,39 +371,109 @@ mod tests {
...
+                }),
+            )
             .unwrap();
 
         (signed_manager_contract, unsigned_manager_contract)
     }
 
     #[test]
     fn call_remote() {
+        let owner = "owner".into_addr();
```
uint, line 402: Yeah, I like this pattern better too :)

sylvia/src/types.rs
```diff
@@ -434,6 +434,19 @@ impl<'a, Contract: ?Sized> Remote<'a, Contract> {
     pub fn executor(&self) -> ExecutorBuilder<(EmptyExecutorBuilderState, Contract)> {
         ExecutorBuilder::<(EmptyExecutorBuilderState, Contract)>::new(&self.addr)
     }
+
+    pub fn update_admin(&self, new_admin: String) -> WasmMsg {
```
uint, line 438: Would a very brief doc comment be good here? I understand this is a public API?

uint (approved):
LGTM! I only wonder if doc comments would be useful with these new pub methods

sylvia/src/types.rs
hashedone, line 438: Consider changing this signature to `pub fn update-admin(&self, new_admin: impl FromStr) -> WasmMsg`. This way it would be more convenient to use with string slices (which you do in tests).  It is, in general, a good habit to avoid `String` as an argument - it is typically better to either go with `&str`/`iml FromStr` (depending if we would make use of ownership).

hashedone (commented):
One comment about API improvement.

sylvia/src/types.rs
jawoznia: Thanks for pointing this out I will add comments ;)

jawoznia: Good point!

sylvia/tests/remote.rs
```diff
@@ -331,39 +371,109 @@ mod tests {
...
 
         let unsigned_manager_contract = manager_code_id
             .instantiate(unsigned_counter_contract.contract_addr.clone())
             .with_label("Manager contract")
-            .call(&OWNER.into_bech32())
+            .call(owner)
+            .unwrap();
+
+        // Set manager contract as an admin of the counter contract
+        app.app_mut()
```
kulikthebird, line 387: Let's create a new setup function for the admin scenarios

kulikthebird (commented):
I like this change, it's nice to have those extra methods in the `Remote` implementation. The only thing I spend more time was checking the `remote.rs` file with tests. I realized it is more complicated than it can be. I know it's not directly related to this PR, but we can do a nice improvement with low cost:

1. `unsigned_contract` and `signed_contract` are pretty much the same - let's remove the code duplication and make a single `Counter` contract generic over the counter type. We can then add `type SignedContract = Counter<i32>; type UnsignedContract = Counter<u32>;`.
2. Names `signed` and `unsigned` reminds me of signing something with some cryptographic tool, when in this context it means `i32` and `u32`. Let's change it to `CounterI32`/`CounterU32` or `ContractI32`/`ContractU32` instead.
3. `ContractStorage` is a dead code - to be removed.
4. Remove the `#[entry_points(generics<i32>)]` - we don't need it in the test scenarios. Moreover it might be misleading for other devs:
 - it's only `i32`, but we can support many other types - better way would be to add feature flags for the contracts to choose between entry points types,
 - it gives a hint that it always should be present with sylvia contracts - it doesn't have to.

jawoznia:
I agree it could be simplified


>     2. Names `signed` and `unsigned` reminds me of signing something with some cryptographic tool, when in this context it means `i32` and `u32`. Let's change it to `CounterI32`/`CounterU32` or `ContractI32`/`ContractU32` instead.

I can see that. Let's change it later.
 
>     3. `ContractStorage` is a dead code - to be removed.

This is intentional. There was a bug where it was impossible to store `Remote` in custom storage types. A comment explaining this would be very much in place.

>     4. Remove the `#[entry_points(generics<i32>)]` - we don't need it in the test scenarios. Moreover it might be misleading for other devs:

It provides the coverage though.
I would prefer to move all the bloated examples to the tests and make them act as the coverage.
The examples should be refactored to provide simple single feature showcases.

PR #440, opened 2024-10-08
feat: Auto deserialize payload

jawoznia:


sylvia/tests/reply.rs
```diff
@@ -198,7 +214,10 @@ where
         &self,
         ctx: ReplyCtx<Q>,
         data: Option<Binary>,
-        _payload: Binary,
+        // Blocked by https://github.com/CosmWasm/cw-multi-test/pull/216.
```
kulikthebird, line 217: Let's add a TODO mark and URL to an issue that will solve this blocked part

kulikthebird (approved):
Looks good! Only one comment here from me. Also it's worth to take a look into `reply.rs` src file before merging this feature into `main`. Maybe there's a chance to make the code more compact in this case

sylvia/tests/reply.rs
jawoznia: Good point. We won't release this feature until fix in MultiTest is released.

I will add `TODO` in following PR as there are other changes to be done.

PR #447, opened 2024-10-15
feat: Add support for `sv::features` attribute. (#446)

### LINKED ISSUE #446: Add `sv::features` attribute to avoid breaking API
To overwrite Sylvia behavior without breaking the API in minor releases we will introduce new attribute `#[sv::features()]`.

User will be able to pass the currently supported feature to enable new behavior.

For the `reply` feature we will support the `#[sv::features(replies)]`.

We will be potentially able to reuse this attribute in the future while implementing better support for `migrate`.

jawoznia:
Remove `sv_replies` feature as it does not make the change non breaking.

closes #446

### DIFF
--- a/sylvia-derive/src/lib.rs
+++ b/sylvia-derive/src/lib.rs
@@ -282,6 +282,7 @@ fn interface_impl(_attr: TokenStream2, item: TokenStream2) -> TokenStream2 {
 ///
 /// ##[sylvia::contract]
 /// ##[sv::error(ContractError)]
+/// ##[sv::features(replies)]
 /// impl SvContract {
 ///     pub const fn new() -> Self {
 ///         Self {
@@ -621,6 +622,13 @@ fn interface_impl(_attr: TokenStream2, item: TokenStream2) -> TokenStream2 {
 /// and only for message types variants that resolves in an enum field,
 /// i.e. `exec`, `query` and `sudo`.
 ///
+/// ### `sv::features(...)`
+///
+/// Enables additional features for the contract. Allows user to use features that
+/// are considered breaking before the major release.
+///
+/// Currently supported features are:
+///     * `replies` - enables better dispatching of `reply` as well as its auto deserialization.
 ///
 #[proc_macro_error]
 #[proc_macro_attribute]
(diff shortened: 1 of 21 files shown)


sylvia/tests/ui/attributes/features/invalid_params.stderr
```diff
@@ -0,0 +1,6 @@
+error: Invalid feature.
+       = note: Expected `#[sv::features(replies)];`.
```
kulikthebird, line 2: It's good to have a list of available features provided in the error message as it is implemented here. Nit: Instead of `Expected [...]` imo we could add `Available features: [feat1, feat2, ...]`. It's worth to note that features attribute probably could be used in few different contexts, so it's important to provide the list of features supported in a given place of use.

sylvia-derive/src/lib.rs
```diff
@@ -621,6 +622,13 @@ fn interface_impl(_attr: TokenStream2, item: TokenStream2) -> TokenStream2 {
 /// and only for message types variants that resolves in an enum field,
 /// i.e. `exec`, `query` and `sudo`.
 ///
+/// ### `sv::features(...)`
+///
+/// Enables additional features for the contract. Allows user to use features that
+/// are considered breaking before the major release.
+///
+/// Currently supported features are:
+///     * `replies` - enables better dispatching of `reply` as well as its auto deserialization.
```
kulikthebird, line 631: That's a very humble description of a really useful and rich feature. I think of adding some URL to the docs at least, or ref to the rustdoc that broader explains replies.

kulikthebird (approved):
I like this approach!

sylvia-derive/src/lib.rs
jawoznia: The docs are not yet there. I will try to expand the description though.

sylvia/tests/ui/attributes/features/invalid_params.stderr
jawoznia: Good point. Especially that this error is regarding the features not the attribute.

PR #449, opened 2024-10-22
chore: Rename `ReplyOn::Failure` to match variant from std

jawoznia:


hashedone (approved):
LGTM, important change.

PR #451, opened 2024-10-22
feat: Add temporary `sylvia::replies::ReplyCtx` with additional fields (#450)

jawoznia:


sylvia/src/types.rs
```diff
@@ -462,12 +460,13 @@ impl<'a, Contract: ?Sized> AsRef<cosmwasm_std::Addr> for Remote<'a, Contract> {
 
 /// Represantation of `reply` context received in entry point.
 #[non_exhaustive]
+#[deprecated(
+    since = "1.3.0",
+    note = "This type will be replaced with `sylvia::replies::ReplyCtx` in 2.0.0."
```
hashedone, line 465: :+1:

hashedone: Only thing I am wondering right now - do we want to deprecate this struct or do we want to throw a deprecation warning using `#[sv::msg(reply)]` without `#[sv::features(replies)]`? I'm not debating on if this is ok to deprecate this, but which one gives back a better warning message (or even error using the wrong context - however I think that the old context could be in fact used with new messages if the proper `From` is implemented? Or not? Or maybe both could be? Let's chat about it).

hashedone (commented):
Mostly ok, but I'd like to quickly talk about deprecation order (it's prob right, but I want to double check)

kulikthebird (commented):
The only comment from me is the one we discussed before - there should be exactly one `ReplyCtx` that is not deprecated

sylvia/src/types.rs
jawoznia: It might be used if we initialize the context with empty vectors and set the gas to 0. I'm not sure if the detection which `ReplyCtx` is used would be possible though.

I went with different approach and moved all the context types to new ctx module.

kulikthebird (approved):
LGTM!

PR #459, opened 2024-11-19
feat: Assert no attributes used on `self` and `ctx`

jawoznia:


sylvia-derive/src/parser/mod.rs
```diff
@@ -67,6 +67,35 @@ pub fn process_fields<'s, Generic>(
 where
     Generic: GetPath + PartialEq,
 {
+    // Assert no accidental usage of Sylvia attributes on `self` and `ctx` parameters.
```
kulikthebird, line 70: Nitpick: consider placing this predicate in a separate function for clarity

kulikthebird (approved):
+1 and one comment

PR #469, opened 2024-12-19
chore: Update cosmwasm to version 2.2

jawoznia:
Since MultiTest is not yet released with a new version, `cosmwasm_2_2` flag in Sylvia won't enable the respective flag on MultiTest side.


kulikthebird (approved):
:+1:  and we should consider adding a new issue for the multitest feature flag addition

DariuszDepta (commented):
LGTM

PR #485, opened 2025-05-20
test: Use original runtime-macros crate

jawoznia:


DariuszDepta:
Will be applied on main.

